Issue #33, opened 2021-11-30
Passing Path object to shutils.move in gdk component init


sgreen-GC:
Hi,

Just been checking the GDK out and I came up against a problem when init'ing a new python component from the HelloWorld template.
See below for reproduction and error:
```
>gdk component init -l python -t HelloWorld
[2021-11-30 12:28:07] INFO - Initializing the project directory with a python component template - 'HelloWorld'.
[2021-11-30 12:28:07] INFO - Fetching the component template 'HelloWorld-python' from Greengrass Software Catalog.

=============================== ERROR ===============================
Could not initialize the project with component template 'HelloWorld'.
'WindowsPath' object has no attribute 'rstrip'
```

I had a look around and it looks like its a problem with passing a Path object to the shutil.move() function:
https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/7750b60dcc2714264f8f6a99c720b055f6eb7bfd/gdk/commands/component/init.py#L110

I changed it to convert the Path to a string and it seemed to work.

I was running on Windows 10, using GDK v1.0.0, in a python3.8.2 virtualenv.
I also tried it on Ubuntu 20.04, under WSL2, and got a similar error:
```
'PosixPath' object has no attribute 'rstrip'
```

Let me know if there's any other detail I can provide.

pelder-lumenix:
Similar problem on Ubuntu 20.04. Tried with python 3.6, 3.7 and 3.8. Likely cause is the assumption that `shutil.move` accepts Path objects. According to [this StackOverflow](https://stackoverflow.com/questions/61327385/trying-to-use-on-path-object-python), it does not. The fix appears to be as simple as:

```
shutil.move(str(f), str(utils.current_directory))
```

pallavi03:
Thanks @sgreen-GC & @pelder-lumenix  for pointing this out. We are waiting for [PR](https://github.com/aws-greengrass/aws-greengrass-gdk-cli/pull/35) authored by Paul to be ready for merging.

pelder-lumenix:
@pallavi03 : The test failure is related to changed arguments passed to the test's mock of `shutil.move`. But, before I change the test, let's take a more detailed look at the failure.

* the message received is that a Path object (either PosixPath or WindowsPath) does not have an `rstrip()` method.
[...]

PR #35, opened 2021-11-30
Fix failure in shutil.move

pelder-lumenix:
**Issue #, if available:** #33

**Description of changes:**
Ensure arguments to `shutil.move` are strings.

Failure on `gdk component init`

Manually
```
gdk component init -t HelloWorld -l python
```

**Checklist:**

### DIFF
--- a/gdk/commands/component/init.py
+++ b/gdk/commands/component/init.py
@@ -107,7 +107,7 @@ def download_and_clean(comp_name, comp_type):
             zfile.extractall(tmpdirname)
             # Moves the unarchived contents from temporary folder (downloaded-zip-folder) to current directory.
             for f in Path(tmpdirname).joinpath(zfile.namelist()[0]).iterdir():
-                shutil.move(str(f), str(utils.current_directory))
+                shutil.move(str(f), utils.current_directory)
 
 
 def get_download_url(comp_name, comp_type):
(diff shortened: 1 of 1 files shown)


pallavi03:
Thanks @pelder-lumenix for the fix. We appreciate it. I executed the CI workflow and it seems to be failing because a unit test is failing. It should be a small fix in [assert statement](https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/main/tests/gdk/commands/component/test_init.py#L191) . Let us know if you can make that change and we can merge it in. Otherwise, we can send another PR for it.

pelder-lumenix:
A more detailed explanation of the second commit:

- failure indicates a Path object (either PosixPath or WindowsPath) does not have an `rstrip()` method.
- in the Python 3.6 shutil implementation, there is exactly one call to `rstrip()` in `_basename()`. See [code](https://github.com/python/cpython/blob/2c56c97f015a7ea81719615ddcf3c745fba5b4f3/Lib/shutil.py#L546).
- In turn, `_basename()` is only called from `shutil.move()`, and only on the `src` (1st) argument. See [code](https://github.com/python/cpython/blob/2c56c97f015a7ea81719615ddcf3c745fba5b4f3/Lib/shutil.py#L510).
- Thus, a minimal fix would be to ensure that the first argument is a string.
- the test failure occurs because the test's mocked `shutil.move()` implementation expects the second argument to be a Path object. The original commit had unnecessarily converted it to a string.

philcali (approved):

philcali:
Thanks for this PR, @pelder-lumenix!

MikeDombo (approved):

PR #36, opened 2021-12-01
Revert "Fix failure in shutil.move"

saranyailla:
Reverts aws-greengrass/aws-greengrass-gdk-cli#35


MikeDombo:
Please explain why this needs to be reverted

PR #48, opened 2022-01-11
refactor(publish): build before publish

saranyailla:
**Issue #, if available:**

**Description of changes:**
If the component artifacts are not build, publish command exits displaying a log. With this change, component builds before publishing if it is not already build. Otherwise, the command directly publishes the component. 

Customers can now create component versions in one command instead of 2 if the component is not build already. 

Updated old tests and added new ones.

**Checklist:**

### DIFF
--- a/gdk/commands/component/publish.py
+++ b/gdk/commands/component/publish.py
@@ -2,6 +2,7 @@ import json
 import logging
 from pathlib import Path
 
+import gdk.commands.component.component as component
 import gdk.commands.component.project_utils as project_utils
 import gdk.common.exceptions.error_messages as error_messages
 import gdk.common.utils as utils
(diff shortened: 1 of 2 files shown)


gdk/commands/component/publish.py
```diff
@@ -11,15 +12,21 @@
...
         project_config["account_number"] = get_account_number()
         project_config["bucket"] = "{}-{}-{}".format(
             project_config["bucket"], project_config["region"], project_config["account_number"]
         )
 
         component_name = project_config["component_name"]
-        logging.info(f"Publishing the component '{component_name}' with the given project configuration.")
         component_version = get_component_version_from_config()
-
+        logging.debug(f"Checking if the component '{component_name}' is build.")
```
pallavi03, line 23: typo: `is built`. Correct other places as well

pallavi03 (approved):

popanmol (approved):

PR #49, opened 2022-01-14
feat(bucket): pass bucket name in publish command

saranyailla:
**Issue #, if available:**

**Description of changes:**
Publish command now takes in additional argument (optional) for the bucket name. 

Using gdk-config.json file, customers can only provide bucket prefix to which account and region are appended. To provide customers with the ability to use the bucket of their choice, adding this optional argument for passing in exact bucket name.  

Updated unit tests. 

**Checklist:**

### DIFF
--- a/gdk/commands/component/publish.py
+++ b/gdk/commands/component/publish.py
@@ -10,13 +10,16 @@ import yaml
 from botocore.exceptions import ClientError
 
 
-def run(args):
+def run(command_args):
     try:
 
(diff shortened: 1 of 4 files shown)


pallavi03 (commented):
We should track public documentation change for this. Also release this as new minor version of GDK along with other changes in command that we have coming up

pallavi03 (approved):

philcali (approved):

PR #55, opened 2022-01-25
chore(version): update gdk version to 1.1.0-dev

saranyailla:
**Issue #, if available:**

**Description of changes:**
Update `gdk` to `1.1.0-dev` as the next release is 1.1.0

Before parsing every command, tool checks for latest available version on github and recommends user to switch to the latest. 
This limits us to not update the gdk version until the day of its release. This also means that the code on main branch with ongoing updates is still identified as its previous version. So, to improve this state, we append "-dev" to the upcoming version and recommend customers to upgrade only when it is not "-dev" type version.

Updated unit tests. Removed integration as it is not correct.

**Checklist:**


saranyailla:
Closing this PR for now as we want to maintain a separate branch for development. Main branch will always contain the latest and stable version of the CLI.

PR #56, opened 2022-01-25
chore(workflow): Adding CI step to ensure that gdk is installed successfully

pallavi03:
**Issue #, if available:**
Recently there was breaking change which impacted cli installation. This was not caught in CI step of integration because dependencies in `test-requirement.txt` were different than actual cli dependencies in  `requirement.txt`. One option is to ensure that both list of dependencies are same, but that means we install test dependencies during runtime which is unnecessary. 

**Description of changes:**
This fix is alternative approach to ensure that we install GDK as customer would and verify that its installed correctly.

E.g. workflow failure if GDK is not installed correctly: https://github.com/aws-greengrass/aws-greengrass-gdk-cli/actions/runs/1748076300

E.g workflow successful  if GDK is installed correctly: https://github.com/aws-greengrass/aws-greengrass-gdk-cli/actions/runs/1748090819

Its necessary to caught installation issue proactively.

Tested with setting up workflow in another branch

**Checklist:**
- [X] You confirm that the change is backwards compatible

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -28,19 +28,21 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install dependencies
+      - name: Install CLI and verify
         run: |
           python -m pip install --upgrade pip
+          pip install .
+          gdk --help
+      - name: Testing CLI (Runs both unit and integration tests)
+        run: |
           if ( Test-Path test-requirements.txt ) { pip install -r test-requirements.txt; }
+          coverage run --source=gdk -m pytest -v -s . && coverage report --show-missing --fail-under=70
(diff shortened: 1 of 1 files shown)


.github/workflows/CI.yml
```diff
@@ -97,4 +97,30 @@ jobs:
       - name: Testing CLI (Runs both unit and integration tests)
         run: |
           pip install .
```
saranyailla, line 99: Can we move installing `test-requirements` to under `Testing` step here and remove installing it  from `Install dependencies` step instead of adding new task for it?

pallavi03: We can but I kept it that way to separate the output. Changed it though

saranyailla (approved):

avipinku (approved):

PR #57, opened 2022-01-28
feat(s3-uri): support s3 artifacts during build

saranyailla:
**Issue #, if available:**

**Description of changes:**
This change allows gdk to create recipes with artifacts that exist in either s3 or local build folders of gdk.

*Current behavior:* 
We parse artifact URIs in the recipe and check for only those artifacts that have s3:// prefix to avoid uris that are docker images. For the ones with s3:// prefix, we check if the files exist in the build folder. If not, the tool exits with an error. 

To support remote s3 uris,
*New behavior:* With this change, we don't exit immediately but check for the artifact in s3 as well. Thus, supporting both local build artifacts and s3 uris in the recipe. 

Currently, we only support local artifacts that are built using gdk build system. There may be cases where some artifacts already exist in customer's s3 buckets and recipes point to them. Eg. It saves more time to refer big ML models that exist on s3 than downloading them locally, and then uploading them back to s3 only for it to be compatible with gdk.

Added unittests

**Checklist:**

### DIFF
--- a/gdk/commands/component/build.py
+++ b/gdk/commands/component/build.py
@@ -276,14 +276,12 @@ def copy_artifacts_and_update_uris():
     logging.info("Copying over the build artifacts to the greengrass component artifacts build folder.")
     logging.info("Updating artifact URIs in the recipe.")
     parsed_component_recipe = project_config["parsed_component_recipe"]
-    gg_build_component_artifacts_dir = project_config["gg_build_component_artifacts_dir"]
-    artifact_uri = "s3://BUCKET_NAME/COMPONENT_NAME/COMPONENT_VERSION"
     if "Manifests" not in parsed_component_recipe:
         logging.debug("No 'Manifests' key in the recipe.")
         return
-
     build_folders = _get_build_folder_by_build_system()
     manifests = parsed_component_recipe["Manifests"]
+    s3_client = None
     for manifest in manifests:
         if "Artifacts" not in manifest:
             logging.debug("No 'Artifacts' key in the recipe manifest.")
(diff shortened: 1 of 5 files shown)


gdk/commands/component/build.py
```diff
@@ -317,13 +317,26 @@ def copy_artifacts_and_update_uris():
...
-                    f"Could not find the artifact file specified in the recipe '{artifact_file}' inside the build folders."
-                    f" '{build_folders}'."
+                    f"Could not find artifact with URI '{artifact_uri}' on s3 or inside the build folders '{build_folders}'."
                 )
 
 
+def is_artifact_in_s3(artifact_uri):
+    bucket_name, object_key = artifact_uri.replace(utils.s3_prefix, "").split("/", 1)
+    try:
+        s3_client = project_utils.create_s3_client(project_config["region"])
```
devnaruka, line 331: How costly is client creation? Is it cached?

saranyailla: Good catch. Right now, client is created for every artifact that's not found in the local build folder. Need to revise the code.

gdk/commands/component/build.py
```diff
@@ -317,13 +317,26 @@ def copy_artifacts_and_update_uris():
...
+        response = s3_client.head_object(Bucket=bucket_name, Key=object_key)
+        if response["HTTPStatusCode"] == 200:
```
avipinku, line 333: any other status codes we should be noting?

avipinku (approved):

gdk/commands/component/build.py
saranyailla: It must be 200.

pallavi03 (approved):

avipinku (approved):

devnaruka (approved):

PR #68, opened 2022-02-10
fix(gradle): identify build folders properly

saranyailla:
**Issue #, if available:**

**Description of changes:**
- Correctly identifies the build folders for a specific build system. 
- Updated unit tests as needed. 

Instead of recursively finding the intersection of parents directories of build files and build folders in a given project which fails when there are multi-path build folders (a/b/c), we first identify the list of directories with build files in a project. We then iterate through this list and check if build folders also exist within them. 

This change properly identifies the project directories with build file (pom.xml, build.gradle) and build folders (target/, build/libs) recursively. 

- Added UAT for gradle build system with multiple projects. 
 - Added UAT for maven build system with multiple projects. 
 
Test data needed for multi project gradle and multi project maven build systems is downloaded from s3 bucket. 

**Checklist:**

### DIFF
--- a/gdk/static/project_build_system.json
+++ b/gdk/static/project_build_system.json
@@ -19,10 +19,6 @@
             "gradle",
             "build"
         ],
-        "build_command_win": [
-            "gradlew",
-            "build"
-        ],
         "build_folder": [
             "build",
             "libs"
--- a/uat/test_uat_build.py
+++ b/uat/test_uat_build.py
@@ -1,13 +1,16 @@
-import json
 import os
+import shutil
 import subprocess as sp
 from pathlib import Path
+
+import t_utils
 
 
 def test_build_template_zip(change_test_dir):
     # Recipe contains HelloWorld.zip artifact. So, create HelloWorld directory inside temporary directory.
     path_HelloWorld = Path(change_test_dir).joinpath("HelloWorld")
-
+    component_name = "com.example.PythonHelloWorld"
+    region = "us-east-1"
     # Check if init downloads templates with necessary files.
     check_init_template = sp.run(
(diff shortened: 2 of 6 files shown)


gdk/static/project_build_system.json
```diff
@@ -20,7 +20,7 @@
             "build"
         ],
         "build_command_win": [
-            "gradlew",
+            "gradle",
```
pallavi03, line 23: why this change?

uat/test_uat_build.py
```diff
@@ -104,3 +107,75 @@ def test_build_template_maven(change_test_dir):
...
+        path_multi_gradle_project,
+        "zip",
+    )
+    os.remove(Path(change_test_dir).joinpath(zip_file))
+    os.chdir(path_multi_gradle_project)
+    assert Path(path_multi_gradle_project).joinpath("recipe.yaml").resolve().exists()
+    config_file = Path(path_multi_gradle_project).joinpath("gdk-config.json").resolve()
+    assert config_file.exists()
+
+    # Update gdk-config file mandatory field like region.
```
pallavi03, line 135: can we use the util method here to update config file?

gdk/static/project_build_system.json
saranyailla: just `gradle` works fine without using `.bat` wrapper

uat/test_uat_build.py
saranyailla: util method had extra fields which are not needed here. But updating it for consistency.

pallavi03 (approved):

devnaruka (approved):
LGTM

NIT: I would encourage to put given, when and then in comments for test cases to better define our intentions. It helps in increasing code readability. It would help in refactoring these test suites in future.

PR #69, opened 2022-02-11
test: add integration tests for list, config, model actions, parse args & project utils

saranyailla:
**Issue #, if available:**

**Description of changes:**
Add integration tests for list, config, model actions, parse args & project utils

**Checklist:**

### DIFF
--- a/gdk/commands/component/list.py
+++ b/gdk/commands/component/list.py
@@ -35,7 +35,7 @@ def get_component_list_from_github(url):
         comp_list = comp_list_response.json()
         return comp_list
     except Exception as e:
-        logging.error(e)
+        logging.error(e, exc_info=True)
         return []
 
 
--- a/integration_tests/gdk/common/test_integ_model_actions.py
+++ b/integration_tests/gdk/common/test_integ_model_actions.py
@@ -40,3 +40,98 @@ def test_cli_model(mocker):
     with open(utils.get_static_file_path(cli_model_schema), "r") as schemaFile:
         schema = json.loads(schemaFile.read())
     jsonschema.validate(cli_data, schema)
+
+
+def test_is_valid_model_without_help_in_command():
+    # Invalid model without help for commands
+    valid_model = {
+        "gdk": {"sub-commands": ["component"]},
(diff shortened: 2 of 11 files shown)


gdk/commands/component/list.py
```diff
@@ -35,7 +35,7 @@ def get_component_list_from_github(url):
         comp_list = comp_list_response.json()
         return comp_list
     except Exception as e:
-        logging.error(e)
+        logging.error(e, exc_info=True)
```
avipinku, line 38: I think it may be redundant to use (e, exc_info)

integration_tests/gdk/common/test_integ_model_actions.py
```diff
@@ -40,3 +40,110 @@ def test_cli_model(mocker):
...
+            "sub-commands": ["component"],
+            "arguments": {"lang": {"names": ["-l", "--lang"], "help": "help"}},
+            "help": "help",
+        },
+        "component": {"help": "help"},
+    }
+    assert not model_actions.is_valid_model(invalid_model_without_name_in_args, consts.cli_tool_name)
+
+
+# def test_is_valid_model_without_help():
```
avipinku, line 69: any reason this is commented out?

avipinku (approved):

gdk/commands/component/list.py
saranyailla: That's right but in other places where we just log with `e`,  we raise meaningful custom exception messages. But, here, we just proceed even when the exception occurs.  

So, this exc_info helps us log when and where details of it in the cases where logs with just `e` produces incomplete messages like this `TypeError: 'str' object is not callable`

integration_tests/gdk/common/test_integ_model_actions.py
saranyailla: Cleaned.

avipinku (approved):

pallavi03 (approved):

avipinku (approved):

pallavi03 (approved):

PR #71, opened 2022-02-14
Cmd refactor

saranyailla:
**Issue #, if available:**

**Description of changes:**
Refactor code to use take Object oriented approach along with exception handling. 

- Now each command is represented as a class. 
- Command base class includes argument validation as part of command instantiation before running the command. 
- All the CLI commands are inherited from this `Command` base class. Each command implements the `run` method in its own way. 

All the changes are backward compatible with minor changes in logging content. 

**Checklist:**


gdk/commands/component/InitCommand.py
```diff
@@ -0,0 +1,135 @@
...
+import gdk.commands.component.list as list
+import gdk.common.consts as consts
+import gdk.common.exceptions.error_messages as error_messages
+import gdk.common.utils as utils
+import requests
+from gdk.commands.Command import Command
+
+
+class InitCommand(Command):
+    def __init__(self, command_args, level_command) -> None:
```
pallavi03, line 17: why `level_command` is required as arg here? `InitCommand` should always have "Init" as `level_command`. So instead we can call super's init with "Init" here directly

pallavi03 (dismissed):

gdk/commands/component/InitCommand.py
```diff
@@ -0,0 +1,135 @@
...
+            raise Exception("Could not initialize the project with component template '{}'.\n{}".format(template, e))
+
+    def init_with_repository(self, repository, project_dir):
+        try:
+            logging.info("Fetching the component repository '{}' from Greengrass Software Catalog.".format(repository))
+            self.download_and_clean(repository, "repository", project_dir)
+        except Exception as e:
+            raise Exception("Could not initialize the project with component repository '{}'.\n{}".format(repository, e))
+
+    def download_and_clean(self, comp_name, comp_type, project_dir):
```
devnaruka, line 87: This download function is generic and could be move into a utility

gdk/commands/component/InitCommand.py
```diff
@@ -0,0 +1,135 @@
...
+          comp_type(string): Type of the component(template or repository).
+
+        Returns
+        -------
+            None
+        """
+        comp_url = self.get_download_url(comp_name, comp_type)
+
+        download_response = requests.get(comp_url, stream=True)
+        if download_response.status_code != 200:
```
devnaruka, line 105: Is `200` enough or you need to check any other `2XX` status code? Also, how are redirects handled?

gdk/commands/component/InitCommand.py
```diff
@@ -0,0 +1,135 @@
...
+                raise Exception(error_messages.INIT_FAILS_DURING_COMPONENT_DOWNLOAD.format(comp_type))
+
+        logging.debug("Downloading the component {}...".format(comp_type))
+        with zipfile.ZipFile(BytesIO(download_response.content)) as zfile:
+            with tempfile.TemporaryDirectory() as tmpdirname:
+                # Extracts the zip file into temporary directory - /some-temp-dir/downloaded-zip-folder/
+                zfile.extractall(tmpdirname)
+                # Moves the unarchived contents from temporary folder (downloaded-zip-folder) to current directory.
+                for f in Path(tmpdirname).joinpath(zfile.namelist()[0]).iterdir():
+                    shutil.move(str(f), project_dir)
```
devnaruka, line 121: This could throw IO error

tests/gdk/commands/component/test_ListCommand.py
```diff
@@ -0,0 +1,84 @@
...
+        mock_template_list = self.mocker.patch("requests.get", return_value=mock_response)
+        self.mocker.patch.object(ListCommand, "__init__", return_value=None)
+        list = ListCommand({})
+        templates_list = list.get_component_list_from_github("url")
+        assert templates_list == []
+        assert mock_template_list.call_count == 1
+
+    def test_get_component_list_from_github_invalid_url(self):
+        mock_response = self.mocker.Mock(
+            status_code=404, raise_for_status=self.mocker.Mock(side_effect=HTTPError("some error"))
```
devnaruka, line 44: add another test for 500

gdk/commands/component/InitCommand.py
saranyailla: Only `200`. We need full content to download the zip at the [URL in software catalog](https://github.com/aws-greengrass/aws-greengrass-software-catalog/blob/main/cli-components/templates.json) on the github. 

By default, `requests` module in python allows redirects for `GET`.

gdk/commands/component/InitCommand.py
saranyailla: Right but we use it only with the `init` command.

gdk/commands/component/InitCommand.py
saranyailla: We wrapped entire function `download_and_clean` under try-catch. Also, even before trying to download, we checked if the `project_dir` is empty or doesn't exist.

pallavi03 (approved):

gdk/commands/component/BuildCommand.py
```diff
@@ -0,0 +1,425 @@
...
+        Parameters
+        ----------
+            build_system(string): build system specified in the gdk config file
+
+        Returns
+        -------
+            build_command(list): List of build commands of the build system specific to the platform.
+        """
+        platform_sys = platform.system()
+        if platform_sys == "Windows" and "build_command_win" in self.supported_build_sytems[build_system]:
```
avipinku, line 125: can these be consts?

gdk/commands/component/BuildCommand.py
```diff
@@ -0,0 +1,425 @@
...
+
+        Parameters
+        ----------
+            path,names
+
+        Returns
+        -------
+            ignore_list(list): List of files or directories to ignore during zip.
+        """
+        # TODO: Identify individual files in recipe that are not same as zip and exclude them during zip.
```
avipinku, line 219: Should this TODO be here? If so, can we add an issue link?

gdk/commands/component/BuildCommand.py
```diff
@@ -0,0 +1,425 @@
...
+        logging.info("Updating artifact URIs in the recipe.")
+        parsed_component_recipe = self.project_config["parsed_component_recipe"]
+        if "Manifests" not in parsed_component_recipe:
+            logging.debug("No 'Manifests' key in the recipe.")
+            return
+        build_folders = self._get_build_folder_by_build_system()
+        manifests = parsed_component_recipe["Manifests"]
+        s3_client = None
+        for manifest in manifests:
+            if "Artifacts" not in manifest:
```
avipinku, line 311: maybe make these consts as well to have a defined schema somewhere. Or use a dedicated parser

gdk/commands/component/ListCommand.py
```diff
@@ -0,0 +1,51 @@
...
+            self.display_list(li)
+            return
+        raise Exception(error_messages.LIST_WITH_INVALID_ARGS)
+
+    def get_component_list_from_github(self, url):
+        comp_list_response = requests.get(url)
+        if comp_list_response.status_code != 200:
+            try:
+                comp_list_response.raise_for_status()
+            except Exception as e:
```
avipinku, line 33: will this log the error status code/other errors? just making sure

gdk/commands/component/ListCommand.py
```diff
@@ -0,0 +1,51 @@
...
+                raise Exception(error_messages.LISTING_COMPONENTS_FAILED)
+        try:
+            comp_list = comp_list_response.json()
+            return comp_list
+        except Exception as e:
+            logging.error(e, exc_info=True)
+            return []
+
+    def display_list(self, components):
+        # TODO: Add short description of what each component is for?
```
avipinku, line 46: TODO callout here

gdk/commands/component/PublishCommand.py
```diff
@@ -0,0 +1,383 @@
...
+            else:
+                location = {"LocationConstraint": region}
+                self.service_clients["s3_client"].create_bucket(Bucket=bucket, CreateBucketConfiguration=location)
+        except ClientError as e:
+            if e.response["Error"]["Code"] == "BucketAlreadyExists":
+                logging.error("Bucket already exists. Please provide a different name for the bucket in the configuration.")
+            elif e.response["Error"]["Code"] == "BucketAlreadyOwnedByYou":
+                if self.bucket_exists_in_same_region(bucket, region):
+                    logging.info("Not creating an artifacts bucket as it already exists.")
+                    return
```
avipinku, line 114: nit: can just `raise e`

gdk/commands/component/BuildCommand.py
saranyailla: Having a dedicated schema parser here hard as we have to keep up with the nucleus and the cloud way of validating a recipe. Though it is unlikely to change the recipe in a backward incompatible way, but in the case we do, we need to support that through gdk as well. 

So we rely on the recipe validation using the create_component_version api directly. 

Using consts is one way to go.

gdk/commands/component/PublishCommand.py
```diff
@@ -0,0 +1,383 @@
...
+        ----------
+            None
+
+        Returns
+        -------
+            version(string): Version of the component to create.
+        """
+        fallback_version = "1.0.0"
+        logging.debug("Fetching private components from the account.")
+        try:
```
avipinku, line 183: I've noticed that `NEXT_PATCH` doesn't work if an existing version isn't already in GCS. Can we set a baseline for that, e.g. to create 1.0.0?

tests/gdk/commands/component/test_PublishCommand.py
```diff
@@ -0,0 +1,827 @@
...
+
+        component_name = "component_name"
+        component_version = "1.0.0"
+        publish = PublishCommand({})
+        with pytest.raises(Exception) as e:
+            publish.update_and_create_recipe_file(component_name, component_version)
+        assert "as it is not build.\nBuild the component `gdk component build` before publishing it." in e.value.args[0]
+        assert mock_create_publish_recipe.call_count == 0
+
+    def test_update_and_create_recipe_file_uri_not_matches(self):
```
avipinku, line 163: maybe add a test for `NEXT_PATCH` when there are no prior versions

gdk/commands/Command.py
```diff
@@ -0,0 +1,95 @@
...
+        ----------
+          _non_conflicting_args_map(dict): A dictionary object that's formed with argument as a key and
+                              a set of its non-conflicting args as value.
+
+        Returns
+        -------
+          (bool) Returns True if the command arguments conflict. Else False.
+        """
+        command_arg_keys = self._arguments_list(_non_conflicting_args_map)
+        for i in range(len(command_arg_keys)):
```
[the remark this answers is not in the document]
saranyailla: We're using `argparse` as the tool builder. Using  `add_mutually_exclusive_group` of argparse, we cannot add more than one set of conflicting groups. There used to be a library for doing this https://pypi.org/project/ConflictsOptionParser/ but because it is not actively maintained, I wrote our own.

gdk/commands/component/BuildCommand.py
```diff
@@ -0,0 +1,425 @@
...
+        ----------
+            None
+
+        Returns
+        -------
+            build_folder(Path): Path to the build folder created by component build system.
+        """
+        build_system = self.project_config["component_build_config"]["build_system"]
+        build_folder = self.supported_build_sytems[build_system]["build_folder"]
+        if build_system == "gradle":
```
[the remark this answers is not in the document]
saranyailla: Right. I have not worked on these pieces yet in this PR. Didn't want such changes go noticed in this huge PR which is copying stuff.  Will have to individually work on these commands.

tests/gdk/commands/component/test_PublishCommand.py
saranyailla: `test_get_next_version_component_not_exists` checks if 1.0.0 is used as the fallback version when a component with no previous versions is published with `NEXT_PATCH`

gdk/commands/component/PublishCommand.py
saranyailla: It should work. https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/main/gdk/commands/component/publish.py#L186-L191

devnaruka (approved):

gdk/commands/component/BuildCommand.py
avipinku: Well it might also be easier to stay up to date with the Nucleus recipe schema if we have a schema to match, but I see your point

gdk/commands/component/PublishCommand.py
avipinku: Hmm. Maybe I was using an older version.

avipinku (approved):

PR #72, opened 2022-02-17
fix(logging): set logging basic config in gdk module __init__

saranyailla:
**Issue #, if available:**

**Description of changes:**
Set basic config of the root logger in gdk `__init__` to make sure that root logger is configured first with the basic config before using it anywhere. This is because root logger ignores any config set on it after its default values are set unless we use force arg (available onlt in py3.8 or above). 

The basic config set in the current GDK doesn't work on versions < py 3.8 as force=True is not available. So, WARNING is set at as default and the log formatter is also not  applied as expected resulting in logs like  https://github.com/aws-greengrass/aws-greengrass-gdk-cli/runs/5150912788?check_suite_focus=true#step:11:23 instead of https://github.com/aws-greengrass/aws-greengrass-gdk-cli/runs/5225356195?check_suite_focus=true#step:11:33

**Checklist:**

### DIFF
--- a/gdk/__init__.py
+++ b/gdk/__init__.py
@@ -1,0 +1,5 @@
+import logging
+
+import gdk.common.consts as consts
+
+logging.basicConfig(level=logging.INFO, format=consts.log_format, datefmt=consts.date_format)
--- a/gdk/CLIParser.py
+++ b/gdk/CLIParser.py
@@ -192,7 +192,6 @@ def main():
 
 
 try:
-    logging.basicConfig(level=logging.INFO, format=utils.log_format, datefmt="%Y-%m-%d %H:%M:%S", force=True)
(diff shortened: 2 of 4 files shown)


gdk/__init__.py
```diff
@@ -0,0 +1,6 @@
+import logging
+
+log_level = logging.INFO
+log_format = "[%(asctime)s] %(levelname)s - %(message)s"
+date_format = "%Y-%m-%d %H:%M:%S"
```
pallavi03, line 5: Should we move these to constants file ?

saranyailla: Done

pallavi03 (approved):

popanmol (approved):

avipinku (approved):

PR #76, opened 2022-02-22
test(uat): migrate tests to use gdk_cli abstraction

devnaruka:
**Issue #, if available:**

**Description of changes:**
Refactor UATs to emit code coverage metrics if executed in that mode.

To identify metric driven coverage from UAT tests.

Run the test locally in:
1. with installed gdk cli mode
    * `coverage run --source=gdk -m pytest -v -s uat`
1. with code coverage mode
    * `coverage run --source=gdk -m pytest -v -s uat --instrumented`

With existing UATs the code coverage shown is 77%.

**Checklist:**

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -43,10 +43,22 @@ jobs:
           pip3 install flake8
           # The GitHub editor is 127 chars wide
           flake8 . --count --max-complexity=10 --max-line-length=127 --show-source --statistics
-      - name: Testing CLI (Runs both unit and integration tests)
+      - name: Unit Testing CLI
         run: |
           pip install -r test-requirements.txt
-          coverage run --source=gdk -m pytest -v -s tests integration_tests && coverage xml --fail-under=90
+          coverage run --source=gdk -m pytest -v -s tests && coverage xml --fail-under=99
+      - name: Upload tests coverage to Codecov
+        uses: codecov/codecov-action@v2
+        with:
+          flags: unit
+      - name: Integration Testing CLI
+        run: |
+          pip install -r test-requirements.txt
+          coverage run --source=gdk -m pytest -v -s integration_tests && coverage xml --fail-under=90
+      - name: Upload integration tests coverage to Codecov
+        uses: codecov/codecov-action@v2
+        with:
+          flags: integ
       - name: Configure AWS Credentials
         uses: aws-actions/configure-aws-credentials@v1
         with:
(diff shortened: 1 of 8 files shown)


uat/conftest.py
```diff
@@ -1,7 +1,24 @@
 import pytest
 
+from t_setup import GdkProcess, GdkInstrumentedProcess
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--gdkimpl", action="store", default="installed", help="supports two gdk implementations: installed or code"
```
saranyailla, line 8: Use action = "store_true" and remove default. Then the coverage commands can be reduced to 
`coverage run --source=gdk -m pytest -v -s uat --gdkimpl  && coverage xml --fail-under=77` and 
`coverage run --source=gdk -m pytest -v -s uat && coverage xml --fail-under=77`

uat/t_setup.py
```diff
@@ -0,0 +1,60 @@
...
+        args = CLIParser.cli_parser.parse_args(arguments)
+
+        exit_code = 0
+        output = ""
+
+        try:
+            f = io.StringIO()
+            with redirect_stdout(f):
+                parse_args_actions.run_command(args)
+            output = f.getvalue()
```
saranyailla, line 55: just run `parse_args_actions.run_command(args)` inside `try` instead of redirecting the output. Should work the same way.

.github/workflows/CI.yml
```diff
@@ -56,8 +68,13 @@ jobs:
...
+        run: |
           coverage run --source=gdk -m pytest -v -s uat
-      - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v2
\ No newline at end of file
+      - name: Run UATs with code coverage
+        run: |
+          coverage run --source=gdk -m pytest -v -s uat --gdkimpl code && coverage xml --fail-under=78
+      - name: Upload UAT coverage to Codecov
+        uses: codecov/codecov-action@v2
```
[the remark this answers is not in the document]
devnaruka: There are two more uploads with unit and integration test tags. Look after unit and test step execution.

uat/conftest.py
devnaruka: So basically remove the value, the infer from existence of the parameter. Ok!

uat/t_setup.py
devnaruka: I need to capture the output to validate the `gdk component list` output based tests.

saranyailla (approved):

pallavi03 (approved):

PR #77, opened 2022-02-22
refactor(build): use custom exceptions in build

saranyailla:
**Issue #, if available:**

**Description of changes:**
Add custom build exceptions to catch different kinds of errors with the build command. 

All exceptions are raised the same way with a reason and a solution. These are all caught at the top level keeping the exception message less redundant. 

**Checklist:**

### DIFF
--- a/gdk/commands/component/BuildCommand.py
+++ b/gdk/commands/component/BuildCommand.py
@@ -7,10 +7,10 @@ from pathlib import Path
 
 import gdk.commands.component.project_utils as project_utils
 import gdk.common.consts as consts
-import gdk.common.exceptions.error_messages as error_messages
 import gdk.common.utils as utils
 import yaml
 from gdk.commands.Command import Command
+from gdk.common.exceptions.BuildError import ArtifactNotFoundException, BuildSystemException, ZipBuildRecursionException
 
 
 class BuildCommand(Command):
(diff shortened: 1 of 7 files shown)


gdk/common/exceptions/BuildError.py
```diff
@@ -0,0 +1,28 @@
+class BuildException(Exception):
```
pallavi03, line 1: Whats the idea behind creating internal exception classes? I understand it helps in testing but we should have guideline on when to create a new exception class, otherwise this practice could be overused.

saranyailla: It not only segregates the error messages based on command type and provide command specific help, it also provides better formatting removing redundant error strings by catching all of them at top level. For example, the solution provided in some of these exceptions are specific to the build command. 
 
I totally agree with the problem of overusing them. But if we expect a possible exception and want to provide customers with very a specific solution or hint for it or even exit, we'd want to create  an exception for it.

pallavi03: As per our offline discussion, we agree to add new exception if we want client to handle it differently. Lets follow this as a practice

popanmol (approved):

gdk/commands/component/BuildCommand.py
```diff
@@ -97,17 +107,14 @@ def default_build_component(self):
...
-            self.find_artifacts_and_update_uri()
+        # From the recipe, copy necessary artifacts (depends on build system) to the build folder .
+        self.find_artifacts_and_update_uri()
 
-            # Update recipe file with component configuration from project config file.
-            self.create_build_recipe_file()
-        except Exception as e:
-            raise Exception("""{}\n{}""".format(error_messages.BUILD_FAILED, e))
+        # Update recipe file with component configuration from project config file.
+        self.create_build_recipe_file()
```
avipinku, line 117: does this no longer raise an exception?

gdk/commands/component/BuildCommand.py
```diff
@@ -142,21 +149,14 @@ def run_build_command(self):
...
-                logging.info("Zipping source code files of the component.")
-                self._build_system_zip()
-            else:
-                logging.info("Running the build command '{}'".format(" ".join(build_command)))
-                sp.run(build_command)
-        except Exception as e:
-            raise Exception(f"Error building the component with the given build system.\n{e}")
+        build_system = self.project_config["component_build_config"]["build_system"]
+        build_command = self.get_build_cmd_from_platform(build_system)
+        if build_system == "zip":
```
avipinku, line 154: nit: this should be a const/abstracted out

avipinku (approved):

PR #78, opened 2022-02-23
refactor(exceptions): use custom exceptions in init & publish

saranyailla:
**Issue #, if available:**

**Description of changes:**
Add custom exceptions in `init` command. Remove error_messages. 

**Checklist:**

### DIFF
--- a/gdk/common/exceptions/InitError.py
+++ b/gdk/common/exceptions/InitError.py
@@ -1,0 +1,37 @@
+class InitException(Exception):
+    def __init__(self, reason="", solution="", exception=None, help_text=""):
+        err_details = ""
+        if exception:
+            err_details = f"Error details: {exception}\n"
+        self.message = f"{reason}\n{err_details}{solution}{help_text}"
+        super().__init__(self.message)
+
+
+class InvalidInitArgumentsException(InitException):
+    def __init__(self, exception=None):
+        reason = "The arguments passed with the command are invalid."
+        solution = "Please initialize the project with correct arguments."
+        help_text = "\n\nTry `gdk component init --help`"
+        super().__init__(reason, solution, exception, help_text)
+
+
+class DirectoryNotEmptyException(InitException):
+    def __init__(self, exception=None):
+        reason = "The current directory is not empty."
+        solution = "Please initialize the project in an empty directory."
+        super().__init__(reason, solution, exception)
+
+
+class InvalidDirectoryException(InitException):
+    def __init__(self, name, exception=None):
+        reason = f"Directory or a file named '{name}' already exsits in the current directory."
+        solution = "Please initialize the project using a different directory name."
+        super().__init__(reason, solution, exception)
+
+
+class ComponentUnavailableException(InitException):
+    def __init__(self, comp_type, comp_name, exception=None):
+        reason = f"Could not find the component {comp_type} '{comp_name}' in Greengrass Software Catalog."
+        solution = "Please specify a valid component name from the Greengrass Software Catalog."
+        help_text = f"\n\nTry `gdk component list --{comp_type}`"
+        super().__init__(reason, solution, exception, help_text)
--- a/gdk/common/exceptions/PublishError.py
+++ b/gdk/common/exceptions/PublishError.py
@@ -1,0 +1,44 @@
+class PublishException(Exception):
+    def __init__(self, reason="", solution="", exception=None):
+        err_details = ""
+        if exception:
+            err_details = f"Error details: {exception}\n"
+        self.message = f"{reason}\n{err_details}{solution}"
+        super().__init__(self.message)
+
+
+class ArtifactNotFoundDuringPublishException(PublishException):
+    def __init__(self, artifact_file, gg_build_component_artifacts, exception=None):
+        reason = (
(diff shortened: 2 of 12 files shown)


gdk/common/exceptions/InitError.py
```diff
@@ -0,0 +1,37 @@
+class InitException(Exception):
+    def __init__(self, reason="", solution="", exception=None, help_text=""):
+        err_details = ""
+        if exception:
+            err_details = f"Error details: {exception}\n"
+        self.message = f"{reason}\n{err_details}{solution}{help_text}"
+        super().__init__(self.message)
+
+
+class InvalidArgumentsException(InitException):
```
pallavi03, line 10: This could be reused for other commands as well

popanmol (dismissed):

gdk/common/exceptions/InitError.py
```diff
@@ -0,0 +1,37 @@
+class InitException(Exception):
+    def __init__(self, reason="", solution="", exception=None, help_text=""):
+        err_details = ""
+        if exception:
+            err_details = f"Error details: {exception}\n"
+        self.message = f"{reason}\n{err_details}{solution}{help_text}"
+        super().__init__(self.message)
```
devnaruka, line 7: This should be generic GDK cli base exception like`GdkException`. Move to a common generic errors file.

If you still want to keep `InitException` for init command specific exception handling during `InitCommand` execution, then you can create a new class `InitException(GdkException)`.

gdk/common/exceptions/InitError.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+class InvalidInitArgumentsException(InitException):
+    def __init__(self, exception=None):
+        reason = "The arguments passed with the command are invalid."
+        solution = "Please initialize the project with correct arguments."
+        help_text = "\n\nTry `gdk component init --help`"
+        super().__init__(reason, solution, exception, help_text)
```
devnaruka, line 15: This can inherit the generic base exception instead. 

The solution and help text could come as arguments as those seem to be specific here. I would suggest to make an overloaded constructor with generic defaults as well.

gdk/common/exceptions/InitError.py
```diff
@@ -0,0 +1,37 @@
...
+        reason = "The current directory is not empty."
+        solution = "Please initialize the project in an empty directory."
+        super().__init__(reason, solution, exception)
+
+
+class InvalidDirectoryException(InitException):
+    def __init__(self, name, exception=None):
+        reason = f"Directory or a file named '{name}' already exsits in the current directory."
+        solution = "Please initialize the project using a different directory name."
+        super().__init__(reason, solution, exception)
```
devnaruka, line 29: This is similar to `InvalidArgumentsException` so it should inherit it. The reason & solution are more contextual.
Also I think this should be renamed to `InvalidNameException`.

gdk/common/exceptions/InitError.py
saranyailla: This has `init` specific error reason and solution. Just renamed it for now. This is more like incomplete args passed to the command. 

This can be refactored but invalid args are actually caught by argparser. 

`gdk component init` is valid for argparser by not for gdk as we expect either template or repo args.

gdk/common/exceptions/PublishError.py
```diff
@@ -0,0 +1,44 @@
+class PublishException(Exception):
+    def __init__(self, reason="", solution="", exception=None):
+        err_details = ""
+        if exception:
+            err_details = f"Error details: {exception}\n"
+        self.message = f"{reason}\n{err_details}{solution}"
+        super().__init__(self.message)
```
devnaruka, line 7: Similar base exception comment applies here too.

gdk/common/exceptions/PublishError.py
```diff
@@ -0,0 +1,44 @@
...
+        solution = "Please check the artifact URI in the recipe and publish the component again."
+        super().__init__(reason, solution, exception)
+
+
+class ComponentNotBuildException(PublishException):
+    def __init__(self, component_name, exception=None):
+        reason = f"The component '{component_name}' is not built."
+
+        solution = "Please build the component using `gdk component build` before publishing it."
+        super().__init__(reason, solution, exception)
```
devnaruka, line 25: This feels more like `InvalidStateException`.

devnaruka (changes requested):
I believe we are creating too many exceptions. For each error message we don't need an exception. Exception hierarchy should be designed based on different exception handling strategies and error messages/solutions should be managed separately in a `Constants` file somewhere. Depending on where these are raised the correct message could be generated and passed to the exceptions.

I would suggest to take a step back and re-think this.

gdk/common/exceptions/InitError.py
saranyailla: Right, worked on those exact changes but this will also go in build command exception. Just wanted to introduce this change in next PR to keep it simple.

gdk/common/exceptions/InitError.py
pallavi03: error reason could be taken as argument to exception, so we dont need to create new exception if message needs to be customized. If exception needs to be handled same way, it should be reused.

PR #85, opened 2022-02-25
chore: add python versions for regression test

devnaruka:
**Issue #, if available:**

**Description of changes:**
Add multiple python versions to verify regression between python versions.

With workflows runs as part of this pull request.

**Checklist:**

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6]
+        python-version: [3.6, 3.7, 3.8, 3.x]
         os: [windows-latest, ubuntu-latest, macos-latest]
     permissions:
       id-token: write
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -13,12 +13,16 @@ on:
   workflow_dispatch:
 
 jobs:
+
+  # Build job focuses on maintaining quality bar of code contributed in each commit.
+  # Enforce static analysis, commit format and code coverage rules etc. here.
+  # To perform any testing on gdk-cli tool, please add them to the regression job.
   build:
-    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.x]
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        python-version: [ 3.7 ]
+        os: [ windows-latest, ubuntu-latest, macos-latest ]
+    runs-on: ${{ matrix.os }}
     permissions:
       id-token: write
       contents: read
@@ -27,19 +31,17 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - name: Validate commit messages 
+      - name: Validate commit messages
         uses: wagoid/commitlint-github-action@v4
         if: matrix.os == 'ubuntu-latest'
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install CLI and verify
+      - name: Python version
         run: |
           python -VV
           python -m pip install --upgrade pip
-          pip install .
-          gdk --help
       - name: Lint with flake8
         run: |
           pip3 install flake8
@@ -48,7 +50,7 @@ jobs:
       - name: Unit Testing CLI
         run: |
           pip install -r test-requirements.txt
-          coverage run --source=gdk -m pytest -v -s tests && coverage xml --fail-under=99
+          coverage run --source=gdk -m pytest -v -s tests && coverage xml --fail-under=98
       - name: Upload tests coverage to Codecov
         uses: codecov/codecov-action@v2
         with:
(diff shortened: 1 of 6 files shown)


.github/workflows/CI.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.6]
+        python-version: [3.6, 3.7, 3.8, 3.x]
```
saranyailla, line 20: Now I wonder if we should include 3.6 as it reached end of life in Dec '21 -  https://endoflife.date/python

.github/workflows/CI.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     runs-on: ${{ matrix.os }}
```
saranyailla, line 17: Also include py version.

.github/workflows/CI.yml
devnaruka: I think we can continue to support for sometime, then discontinue support.

.github/workflows/CI.yml
devnaruka: `runs-on` is only for machine types (refer [this](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on)). 

Python setup step is where python version matters and it also print the python version to ease out the debugging to figure which python version was used for that run.

```
      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-python@v2
        with:
          python-version: ${{ matrix.python-version }}
```

Not really sure what you meant by the comment. Let me know if you still see something missing.

saranyailla: Commented at wrong line. This comment is applicable for the https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/39b8bfd9d61e51f90eedb1e91fb23e633d1b5c60/.github/workflows/CI.yml#L25

integration_tests/gdk/components/test_integ_BuildCommand.py
```diff
@@ -148,7 +147,6 @@ def test_build_run_default_zip_json(mocker, supported_build_system, rglob_build_
         mock_json_dump.call_count == 1
 
     assert mock_get_proj_config.assert_called_once
-    assert not mock_subprocess_run.called
```
saranyailla, line 151: Nit: LGTM or we can assert that maven/gradle build commands are not called using subprocess like this

`assert not mock_subprocess.called_with_args(['mvn','clean','package'])`

devnaruka: But that's a growing list too! Whenever we there is a new build strategy added, this test would need a change, so I refrained from using that.

.github/workflows/CI.yml
devnaruka: Sure .. now that makes sense :)

.github/workflows/CI.yml
```diff
@@ -16,13 +16,14 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
+      max-parallel: 2
```
devnaruka, line 19: I will try to increase the max-parallel value.

devnaruka: I have increased max-parallel to 4. It takes around 30 minutes to complete all workflows.

.github/workflows/CI.yml
```diff
@@ -16,13 +16,14 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     strategy:
+      max-parallel: 6
```
saranyailla, line 19: Since UATs for publish command are failing because two or more tests end up calculating the next version of the component at the same and trying to create exactly the same version (out of which only one would succeed). 

Better way to solve this would be to create `com.python.HelloWorld.somerandomstring`(never exists and fallbacks to use 1.0.0) instead of just `com.python.HelloWorld` to prevent test runs from conflicting and being flakey. Though this solves the problem, we need to still cover the case where we want to calculate the next version of an existing component.

saranyailla (dismissed):

.github/workflows/CI.yml
devnaruka: I agree, this was just placed as a stop-gap solution while I evaluate all publish UAT tests. Until that evaluation and solution, I didn't want workflows to fail.

.github/workflows/CI.yml
devnaruka: This configuration has been removed!

.github/workflows/CI.yml
devnaruka: `max-parallel` config has been removed. As now tests are generating unique component names to avoid name & version conflicts.

saranyailla (approved):

.github/workflows/CI.yml
```diff
@@ -71,9 +72,11 @@ jobs:
           java-version: 1.8
       - name: Run UATs
         run: |
-          coverage run --source=gdk -m pytest -v -s uat
+          # pytest -v -s uat --gdk-debug
```
pallavi03, line 75: do we need to keep it here?

devnaruka: no we don't. Just wanted to keep it handy, in case someone needs it.

.github/workflows/CI.yml
```diff
@@ -69,13 +68,53 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 1.8
-      - name: Run UATs
+      - name: Install CLI and dependencies
         run: |
-          coverage run --source=gdk -m pytest -v -s uat
+          pip install .
+          gdk --help
+          pip install -r test-requirements.txt
```
saranyailla, line 75: Related to previous commit, L56 installs test requirements even before testing CLI tool verification.

.github/workflows/CI.yml
```diff
@@ -14,31 +14,30 @@ on:
...
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-version }}
-      - name: Install CLI and verify
+      - name: Python version
         run: |
+          python -VV
           python -m pip install --upgrade pip
-          pip install .
```
saranyailla, line 40: Why are we making this change when it was added as part of https://github.com/aws-greengrass/aws-greengrass-gdk-cli/pull/56?

.github/workflows/CI.yml
saranyailla: When customers use pip command to install the cli tool, test-requirements are not installed instead the dependencies in the `requirements` file are only installed as part of setup script. So, we want to replicate the same behavior.

.github/workflows/CI.yml
```diff
@@ -69,13 +68,53 @@ jobs:
...
       - name: Run UATs with code coverage
         run: |
           coverage run --source=gdk -m pytest -v -s uat --instrumented && coverage xml --fail-under=77
       - name: Upload UAT coverage to Codecov
         uses: codecov/codecov-action@v2
         with:
-          flags: uat
\ No newline at end of file
+          flags: uat
+  regression:
```
saranyailla, line 83: Should we move this into a separate workflo?

devnaruka: I am going to revisit this anyways with GDK version regression story. I think we are good with this state in interim.

.github/workflows/CI.yml
```diff
@@ -69,13 +68,53 @@ jobs:
...
+    strategy:
+      max-parallel: 3
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.x]
```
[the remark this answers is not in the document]
devnaruka: I would still keep it here. As UATs only run in coverage mode in build to emit metrics.

.github/workflows/CI.yml
devnaruka: I moved the gdk installation closer to UAT step. The installation is not removed.

devnaruka: Synced up offline. The responsibility of verifying clean install (originally identified in #56) is now part of regression job.

.github/workflows/CI.yml
```diff
@@ -69,13 +72,54 @@ jobs:
         uses: actions/setup-java@v1
         with:
           java-version: 1.8
-      - name: Run UATs
-        run: |
-          coverage run --source=gdk -m pytest -v -s uat
       - name: Run UATs with code coverage
         run: |
+          pip install -r test-requirements.txt
```
saranyailla, line 77: Nit: Already installed in L57.

saranyailla (approved):

avipinku (approved):

PR #88, opened 2022-03-01
chore(regression): regression test old gdk version

devnaruka:
**Issue #, if available:**

**Description of changes:**
Add multiple gdk versions to verify regression between latest and old gdk versions.

1. Add new option --target-version to pass the expected the gdk-cli versions to UATs.
   * Example:
      *  `pytest -v -s uat --target-version 1.0.0`
      *  `pytest -v -s uat --target-version 1.1.0`
      *  `pytest -v -s uat --target-version HEAD`
   * If not provided then target version defaults to `HEAD`, which translates to the code at the most recent commit against which the regression workflow is running.
2. Add `@pytest.mark.version(...)` test marker to provide version constraints for any UAT test.
   * The marker takes `kwargs` (key & values) to define constraints. These can be combined to create upper and lower bounds. Available key args are:
      * `eq`: equal to gdk cli version (exact version match).
      * `min` : minimum gdk cli version (lower bound)
      * `ge`: greater than or equal to gdk cli version (lower bound). Alias of `min`.
      * `gt`: greater than gdk cli version (lower bound).
      * `max` : maximum gdk cli version (upper bound)
      * `le`: less than or equal to gdk cli version (upper bound). Alias of `max`.
      * `lt`: less than gdk cli version (upper bound).
   * Examples:
       * `@pytest.mark.version(eq='1.1.0')`
       * `@pytest.mark.version(min='1.1.0')` or `@pytest.mark.version(ge='1.1.0')`
       * `@pytest.mark.version(max='1.1.0')` or `@pytest.mark.version(le='1.1.0')`
       * `@pytest.mark.version(gt='1.1.0')`
       * `@pytest.mark.version(lt='1.1.0')`
       * `@pytest.mark.version(min='1.0.0', max='1.1.0')`
 3. Applied version constraints based on feature changes between versions.

To ensure new changes are not breaking older versions.

With workflows runs as part of this pull request.

**Checklist:**

### DIFF
--- a/uat/README.md
+++ b/uat/README.md
@@ -1,0 +1,103 @@
+# End-to-end tests for Greengrass Development Kit (GDK) - Command Line Interface (CLI)
+
+### *Copyright Amazon.com, Inc. or its affiliates. All Rights Reserved.*
+
+End-to-end setup uses `pytest` and expects test dependencies to be installed.
+
+### Getting Started
+
+#### Prerequisites
+1. [Python3](https://www.python.org/downloads/) and [pip](https://pip.pypa.io/en/latest/installation/): As the GDK CLI tool is written in python, you need to have python3 and pip installed. The most recent version of python includes pip.
+
+2. [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/getting-started-install.html): As you'd have to configure your AWS credentials using AWS CLI before running certain gdk commands.
+
+#### 1. Cloning CLI repository
+
+To install the latest version of CLI using this git repository and pip, run the following command
+
+```shell
+git clone https://github.com/aws-greengrass/aws-greengrass-gdk-cli.git@v1.1.0
+cd aws-greengrass-gdk-cli
+```
+
+#### 2. Install CLI from code
+
+```shell
+python3 -m pip install pip
+pip install .
+```
+
+#### 3. Configure AWS credentials
+
+Configure AWS CLI with your credentials as shown here - https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html.
+
+#### 4. Installing test dependencies
+
+```
+pip install -r test-requirements.txt
+```
+
+#### 5. Running end-to-end test suite
+
+```shell
+pytest -v -s uat
+```
+
+### Built-in options
+
+#### `--gdk-debug` option
+
+Pass this flag to emit debug logs from gdk cli during all test runs. Example:
+```shell
+pytest -v -s uat --gdk-debug
+```
+
+#### 1. `--gdk-debug` option
+
+Pass this flag to emit debug logs from gdk cli during all test runs. Example:
+```shell
+pytest -v -s uat --gdk-debug
+```
+
+#### 2. `--instrumented` option
+
+Pass this flag to in combination with coverage to track code coverage by the test suite.
+```shell
+coverage run --source=gdk -m pytest -v -s uat --instrumented 
+```
+
+Note: The `coverage` is required to instrument the code.
+
+#### 3. `--target-version` option
+Pass this flag to run test suite considering a specific gdk version.
+```shell
+pytest -v -s uat --target-version 1.1.0
+```
+
+If not provided than target version defaults to `HEAD`, which makes version constraint evaluation to assume
+the tests are running against the most recent commit.
+
+#### 4. `@pytest.mark.version(...)` constraints
+`@pytest.mark.version(...)` is pytest marker to enforce version constraints for any test.
+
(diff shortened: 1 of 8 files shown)


uat/test_uat_build.py
```diff
@@ -35,10 +39,40 @@ def test_build_template_zip(change_test_dir, gdk_cli):
         .joinpath("HelloWorld.zip")
         .resolve()
     )
-
     assert artifact_path.exists()
 
 
+def test_build_template_zip_fail_with_no_artifact_since_v1_0_0(change_test_dir, gdk_cli):
```
saranyailla, line 45: instead of adding `since_` here, can we mark it as `min='1.0.0'`?

devnaruka: done

uat/conftest.py
```diff
@@ -21,8 +25,61 @@ def change_test_dir(tmpdir, monkeypatch):
...
+    # request.node is the current test item
+    # query the marker "version" of current test item
+    version_marker = request.node.get_closest_marker('version')
+
+    # if test item was not marked, there's no version restriction
+    if version_marker is None:
+        return
+
+    # parse target version from options
+    target_version = version_parser(version_option) if not latest_mode else version_option
```
saranyailla, line 53: Why would target_version be either of `version_parser(version_option)` and `version_option`?

uat/conftest.py
```diff
@@ -21,8 +25,61 @@ def change_test_dir(tmpdir, monkeypatch):
...
+
+    # if used as @pytest.mark.version(eq=1.0.0)
+    eq_version = version_marker.kwargs.get('eq')
+    eq_version = version_parser(eq_version) if eq_version is not None else None
```
saranyailla, line 57: if `version_parser(None) `is None, then can L56 and 57 be combined?

uat/conftest.py
```diff
@@ -21,8 +25,61 @@ def change_test_dir(tmpdir, monkeypatch):
...
+    if eq_version is not None and (latest_mode or target_version == eq_version):
```
saranyailla, line 58: Here and everywhere, `if x is not None` can be shortened as `if x`. 
Also, if there's a possibility of adding `version(eq='')`accidentally on tests which should be treated as None, `if x` is the way to go unless `version_parser` has a diff behavior with `None` vs `''`.

uat/conftest.py
devnaruka: This simplifies the messages passed to pytest.skip(...) later during the version constraints evaluates.

if its `latest_mode` then `target_version` would evaluate to 'HEAD' string else a `Version` object.

uat/conftest.py
devnaruka: I check for None beforehand as `version_parser(None)` throws error.

pallavi03 (approved):

uat/conftest.py
```diff
@@ -21,8 +25,61 @@ def change_test_dir(tmpdir, monkeypatch):
...
+    version_option = request.config.getoption('--target-version')
+    if version_option is None:
+        return
+
+    latest_mode = version_option == 'HEAD'
+    print(f"\nTarget gdk-cli version {version_option}")
+
+    # request.node is the current test item
+    # query the marker "version" of current test item
+    version_marker = request.node.get_closest_marker('version')
```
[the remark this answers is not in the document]
devnaruka: If both upper and lower bounds provided then, both would be evaluated and enforced.

saranyailla (dismissed):

uat/conftest.py
devnaruka: I like the shortened version. I will update.

uat/README.md
```diff
@@ -0,0 +1,103 @@
...
+```shell
+python3 -m pip install pip
+pip install .
+```
+
+#### 3. Configure AWS credentials
+
+Configure AWS CLI with your credentials as shown here - https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html.
+
+#### 4. Installing test dependencies
```
saranyailla, line 34: IMO, this can be moved to a separate README inside github workflows or uat folder. This is the landing page for customers and it has more information than needed to install and work with gdk. :)

uat/README.md
```diff
@@ -0,0 +1,103 @@
+# End-to-end tests for Greengrass Development Kit (GDK) - Command Line Interface (CLI)
+
```
saranyailla, line 2: Thanks for adding this README!!

saranyailla (approved):

pallavi03 (approved):

PR #89, opened 2022-03-01
fix(publish): warn instead of raising an exception when an artifact i…

saranyailla:
…s not found

**Issue #, if available:**

**Description of changes:**
When an artifact is not found in the build folder during updating the recipe in`publish`,  instead of raising an exception just log it as a warning. This is because that artifact might actually be available on s3 already.   

For the publish command to work as expected with mixed artifacts (Artifacts created locally with build + artifacts on s3).

**Checklist:**

### DIFF
--- a/gdk/commands/component/PublishCommand.py
+++ b/gdk/commands/component/PublishCommand.py
@@ -339,7 +339,7 @@ class PublishCommand(Command):
                     logging.debug("Updating artifact URI of '{}' in the recipe file.".format(artifact_file))
                     artifact["URI"] = f"{artifact_uri}/{artifact_file}"
                 else:
-                    raise Exception(
+                    logging.warning(
                         f"Could not find the artifact file specified in the recipe '{artifact_file}' inside the build folder"
                         f" '{gg_build_component_artifacts}'."
                     )
--- a/tests/gdk/commands/component/test_PublishCommand.py
+++ b/tests/gdk/commands/component/test_PublishCommand.py
@@ -1,3 +1,4 @@
+import logging
 from pathlib import Path
 from unittest import TestCase, mock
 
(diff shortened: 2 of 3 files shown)


avipinku (dismissed):

gdk/commands/component/PublishCommand.py
```diff
@@ -339,7 +339,7 @@ def update_and_create_recipe_file(self, component_name, component_version):
                     logging.debug("Updating artifact URI of '{}' in the recipe file.".format(artifact_file))
                     artifact["URI"] = f"{artifact_uri}/{artifact_file}"
                 else:
-                    raise Exception(
+                    logging.warning(
                         f"Could not find the artifact file specified in the recipe '{artifact_file}' inside the build folder"
                         f" '{gg_build_component_artifacts}'."
                     )
```
devnaruka, line 345: Can we verify if the file is on S3 or not? If its not on S3 then it should be an exception.

devnaruka (changes requested):
I think this can be reproduced in UAT. Let's add a UAT test too.

gdk/commands/component/PublishCommand.py
saranyailla: That check is performed during gdk `build` and will again be with checked by GG's create component version api after the recipe is updated. It is redundant to perform this same check again here. So, just warned saying that it's not in the build folder. 

Check in build: https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/main/gdk/commands/component/BuildCommand.py#L370-L384

Create component after updating recipes: https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/main/gdk/commands/component/PublishCommand.py#L43-L47

devnaruka (approved):

avipinku (approved):

PR #95, opened 2022-03-18
Added support for Gradle’s Kotlin DSL

timmattison:
**Description of changes:**
Add support for Gradle's Kotlin DSL by adding "build.gradle.kts" to the list of build files GDK searches for when the build system is set to "gradle"

Without this change anyone using Gradle's Kotlin DSL would receive an error message like the one below when attempting to run `gdk component build`:

```
[2022-03-17 22:00:20] WARNING - Could not find the artifact file 'HelloWorld-1.0.0.jar' in the build folder 'set()'.
```

I built a Greengrass component that uses Gradle's Kotlin DSL and published it successfully

### DIFF
--- a/gdk/commands/component/BuildCommand.py
+++ b/gdk/commands/component/BuildCommand.py
@@ -247,7 +247,8 @@ class BuildCommand(Command):
         build_system = self.project_config["component_build_config"]["build_system"]
         build_folder = self.supported_build_sytems[build_system]["build_folder"]
         if build_system == "gradle":
-            return self.get_build_folders(build_folder, "build.gradle")
+            return self.get_build_folders(build_folder, "build.gradle")\
+                .union(self.get_build_folders(build_folder, "build.gradle.kts"))
         elif build_system == "maven":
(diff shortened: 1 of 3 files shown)


timmattison:
Additional tests should be added to validate that Kotlin DSL scripts work properly. I can provide those scripts if necessary.

saranyailla:
PR looks good to me. But to proceed further, you need to make the following changes -

1. Update the commit messages to match the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#examples) format for the GitHub workflow checks to pass. You can also refer to the previous commits of this repository.
2. Add relevant unit tests and integration tests to maintain the code coverage of the repository so the codecov can pass.
3. Once you're done with 1 and 2, one of the Greengrass engineers will add UATs (mainly test data) for the same.

Let us know if you need more information.

saranyailla:
Will add UATs in the next PR as GitHub workflow won't run UATs when the PR originates from the external contributor.

Nelsonochoam (approved):

PR #96, opened 2022-03-18
feat: add support for using Gradle wrapper scripts

timmattison:
**Issue:**
https://github.com/aws-greengrass/aws-greengrass-gdk-cli/issues/84

**Description of changes:**
Added support for using Gradle wrapper scripts AKA "./gradlew"

**Why is this change necessary:**
Gradle wrappers enforce that a build uses the proper version of Gradle. Without this change components will be built with the system's Gradle command which may be the wrong version.

**How was this change tested:**
I built a component locally with it

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #97, opened 2022-03-18
Clean up build directories after a failed build

timmattison:
NOTE: `project_config["gg_build_component_artifacts_dir"]` is the absolute path of `greengrass-build/artifacts/COMPONENT_NAME/NEXT_PATCH`
NOTE: `project_config["gg_build_directory"]` is the absolute path of `greengrass-build`

**Description of changes:**
If an exception is thrown during a build the `project_config["gg_build_directory"]` is cleaned up

The current implementation does not check the returncode from Python's subprocess module when it executes a build. This means that if a build fails it still continues to run.

If a user invokes `gdk component publish` on a system that doesn't have an existing build directory and has a broken build the following will happen:

- gdk will create the `project_config["gg_build_directory"]` directory 
- gdk will run the build and it will fail
- gdk will not clean up the `project_config["gg_build_directory"]` directory
- If the last built version of the component is still in the `build/libs` directory (when using Gradle) gdk will publish a new component with the old version without returning a non-zero exit status
- gdk will skip the build process on subsequent publish runs because it sees that `greengrass-build/artifacts/COMPONENT_NAME/NEXT_PATCH` exists

Here is an example log I captured:

```
% gdk component publish

[2022-03-17 23:31:01] INFO - Getting project configuration from gdk-config.json
[2022-03-17 23:31:01] INFO - Found component recipe file 'recipe.yaml' in the  project directory.
[2022-03-17 23:31:01] INFO - Found credentials in shared credentials file: ~/.aws/credentials
[2022-03-17 23:31:01] INFO - Using '1.0.27' as the next version of the component 'com.example.JavaHelloWorld' to create.
[2022-03-17 23:31:01] WARNING - The component 'com.example.JavaHelloWorld' is not built.
So, building the component before publishing it.
[2022-03-17 23:31:02] INFO - Getting project configuration from gdk-config.json
[2022-03-17 23:31:02] INFO - Found component recipe file 'recipe.yaml' in the  project directory.
[2022-03-17 23:31:02] INFO - Building the component 'com.example.JavaHelloWorld' with the given project configuration.
[2022-03-17 23:31:02] INFO - Using 'gradle' build system to build the component.
[2022-03-17 23:31:02] WARNING - This component is identified as using 'gradle' build system. If this is incorrect, please exit and specify custom build command in the 'gdk-config.json'.
[2022-03-17 23:31:02] INFO - Running the build command 'gradle build'

... Gradle build failure omitted ...

BUILD FAILED in 2s
[2022-03-17 23:31:05] INFO - Copying over the build artifacts to the greengrass component artifacts build folder.
[2022-03-17 23:31:05] INFO - Updating artifact URIs in the recipe.
[2022-03-17 23:31:05] INFO - Creating component recipe in '/home/test/HelloWorld/greengrass-build/recipes'.
[2022-03-17 23:31:05] INFO - Publishing the component 'com.example.JavaHelloWorld' with the given project configuration.
[2022-03-17 23:31:05] INFO - Uploading the component built artifacts to s3 bucket.
[2022-03-17 23:31:05] INFO - Uploading component artifacts to S3 bucket: test-greengrass-components-us-east-1-111111111111. If this is your first time using this bucket, add the 's3:GetObject' permission to each core device's token exchange role to allow it to download the component artifacts. For more information, see https://docs.aws.amazon.com/greengrass/v2/developerguide/device-service-role.html.
[2022-03-17 23:31:05] INFO - Successfully created the artifacts bucket 'test-greengrass-components-us-east-1-111111111111' in region 'us-east-1'
[2022-03-17 23:31:05] INFO - Updating the component recipe com.example.JavaHelloWorld-1.0.27.
[2022-03-17 23:31:05] INFO - Creating a new greengrass component com.example.JavaHelloWorld-1.0.27
[2022-03-17 23:31:06] INFO - Created private version '1.0.27' of the component in the account.'com.example.JavaHelloWorld'.

% echo $?
0
```

**To reproduce:**

- Create a working Java component for Greengrass
- Build it with the version set to `NEXT_PATCH`
- Publish it once
- Delete the `project_config["gg_build_component_artifacts_dir"]` directory
- Modify the Java code or Gradle script so it no longer compiles
- Run `gdk component publish`

I broke my build several times and made sure it didn't get published and it tried to rebuild each time.

I have not tested this with the `zip`, `maven`, or `custom` options but if they return errors they'll likely fail in a similar way without this change.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@
 *htmlcov/
 *.coverage
 *.vscode
+venv/
--- a/gdk/build_system/BuildSystem.py
+++ b/gdk/build_system/BuildSystem.py
@@ -1,0 +1,23 @@
+
+class BuildSystem:
+    """
+    Delegates build tasks to the appropriate build system
+    """
+    _instance = None
+
+    def __new__(cls):
+        if cls._instance is None:
+            cls._instance = super(BuildSystem, cls).__new__(cls)
+            cls._instance.systems = {}
+        return cls._instance
+
+    def register(self, system):
+        self.systems[system.__str__()] = system
+
+    def build(self, system_type):
+        system = self.systems.get(system_type)
+
+        if not system:
+            raise TypeError(f"system type {system_type} is not currently supported")
+
+        return system.build()
--- a/gdk/build_system/Zip.py
+++ b/gdk/build_system/Zip.py
@@ -1,0 +1,80 @@
+import shutil
+import logging
+from pathlib import Path
+
+import gdk.common.utils as utils
+import gdk.common.consts as consts
+
+
+class Zip:
+    """
+    Builds the component as a zip file.
+
+    Copies over necessary files by excluding certain files to a build folder identied for zip build system
+    (supported_component_builds.json has the build folder info).
+    This build folder is zipped completely as a component zip artifact.
+    Raises an exception if there's an error in the process of zippings.
+    """
+    def __init__(self, project_config, build_folder):
+        self.build_folder = build_folder
+        self.project_config = project_config
+
+    def __str__(self):
+        return "zip"
+
+    def build(self):
+        self._build_system_zip()
+
+    def _build_system_zip(self):
+        try:
+            zip_build = next(iter(self.build_folder))  # Only one zip-build folder in the set
+            artifacts_zip_build = Path(zip_build).joinpath(utils.current_directory.name).resolve()
+            utils.clean_dir(zip_build)
+            logging.debug("Copying over component files to the '{}' folder.".format(artifacts_zip_build.name))
+            shutil.copytree(utils.current_directory, artifacts_zip_build, ignore=self._ignore_files_during_zip)
+
+            # Get build file name without extension. This will be used as name of the archive.
+            archive_file = utils.current_directory.name
+            logging.debug(
+                "Creating an archive named '{}.zip' in '{}' folder with the files in '{}' folder.".format(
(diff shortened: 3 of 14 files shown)


saranyailla:
Hi Tim, thank you for finding these issues out. I'd like to provide a bit more details about why this is happening and how we can address this in another way.

1. We are not checking the return code of the sub-process while running the build commands. So, if the build fails, we do nothing but proceed. Agree that it needs to be fixed. 
2. Before running any type of build command (zip, maven, custom or even gradle), we clean up `greengrass-build` directory. We also clean additional directories as follows -
   - For **zip**, we also clean up `zip-build` directory.
   - For **maven**, we run `mvn clean package` so that cleans the previously built directory.
   - For **custom**, we leave the responsibility to the customer to clean up the previously built artifacts.   
   - _But for **gradle**, we were not cleaning the built artifacts as we were only using `gradle build` command. This was already caught and addressed by you in the [other PR.](https://github.com/aws-greengrass/aws-greengrass-gdk-cli/pull/96/commits/d366a5043a5d6b377cdef536692779e448c2b915)_ 

For 1, we can add  `check=True` in the sub-process [run command](https://docs.python.org/3/library/subprocess.html#subprocess.run) so it exits when there's an exception in the command. It is similar to what your code change does but this simplifies unit testing. This way, we don't need to clean up the build directory when the build fails. In fact, the intention was to not delete the build folder. 

For 2, we can let that change go in with this PR or the other one, whichever goes first. 

Let us know what you think about this. Thanks again.

PR #99, opened 2022-03-22
added encryption and metadata to artifact upload

JustinKuskowski:
**Issue #, if available:**

**Description of changes:**
This enhancement will allow users to choose an encryption type while uploading the GDK artifact(s) to s3 or if chosen the  code will lookup the default bucket encryption type and use that.  This enhancement also will allow users to add metadata to the artifact(s) that are being uploaded. 

**Supported Encryption Values:** 
_alias_ : Which would use a kms key alias for encrypting the gdk artifact
example 
```
"publish": {
    "bucket": "this-is-a-test",
    "region": "us-east-1",
    "encrypt": {
        "server_side_encryption": "aws:kms",
        "kms_key_id": "alias/gdk-test"
    }
}
```

_kms key arn_ : This option would use a kms key arn for encrypting the gdk artifact
example : 
```
"publish": {
    "bucket": "this-is-a-test",
    "region": "us-east-1",
    "encrypt": {
        "server_side_encryption": "aws:kms",
        "kms_key_id": "arn:aws:kms:us-east-1:###########:key/aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
    }
}
```

_kms key id_ : This option would use a kms key id for encrypting the gdk artifact
example : 
```
"publish": {
    "bucket": "this-is-a-test",
    "region": "us-east-1",
    "encrypt": {
        "server_side_encryption": "aws:kms",
        "kms_key_id": "aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee"
    }
}
```

**MetaData usage:** 
Example : 
```
"publish": {
    "bucket": "this-is-a-test",
    "region": "us-east-1",
    "metadata": {
        "key1": "value1",
        "key2": "value2",
        "key3": "value3"
    }
}
```

A number of company's require a bucket policy to which will deny unencrypted object uploads. The metadata enhancement will make it easier deployment solutions like the Connected Device Framework (CDF) to deploy a GreenGrass component to a device. 

User tested.  I setup a number of scenarios by creating a gdk-config.json file per test.  I also needed to setup S3 bucket default encryption a few different ways for my tests. 

**Checklist:**

### DIFF
--- a/gdk/commands/component/PublishCommand.py
+++ b/gdk/commands/component/PublishCommand.py
@@ -80,7 +80,41 @@ class PublishCommand(Command):
                 }
             return encrypt
         except Exception as e:
-            logging.error("Failed to get default encryption key for S3 Bucket {}".format(bucket_name))
+            logging.error("Error while getting default encryption key for S3 Bucket {}".format(bucket_name))
+            raise Exception("{}\n{}".format(error_messages.PUBLISH_FAILED, e))
+
+    def set_s3_artifact_encryption(self, encrypt, bucket):
+        """
+        Creates S3 Bucket encryption argument
+
+        Parameters
+        ----------
+            encrypt(string | dict): Encryption type and possible KMS Key ID or ARN
+            bucket(string): S3 Bucket Name, used to get default encryption
+
+        Returns
+        -------
+            None
+        """
+        try:
+            extra_args = {}
+            if (type(encrypt) is str) and (encrypt.lower() == 'true'):
+                logging.info("Using default encryption.")
+                extra_args = {'ServerSideEncryption': 'AES256'}
+            elif (type(encrypt) is str) and (encrypt.lower() == 'default'):
+                logging.info("Using default bucket encryption.")
+                extra_args = self.get_bucket_encryption(bucket_name=bucket)
+            elif (type(encrypt) is dict) and encrypt.get('kms_key_id'):
+                logging.info("Using KMS encryption.")
+                extra_args = {
+                    "ServerSideEncryption": encrypt.get('server_side_encryption'),
+                    "SSEKMSKeyId": encrypt.get('kms_key_id')
+                }
+            else:
+                logging.info("No encryption configuration found.")
+            return extra_args
+        except Exception as e:
+            logging.error("Error while setting encryption for S3 Bucket {}".format(bucket))
             raise Exception("{}\n{}".format(error_messages.PUBLISH_FAILED, e))
 
     def upload_artifacts_s3(self, component_name, component_version):
@@ -105,22 +139,7 @@ class PublishCommand(Command):
             metadata = self.project_config.get("metadata")
 
             # Encryption artifact setup
-            if (type(encrypt) is str) and (encrypt.lower() == 'true'):
-                logging.info("Using default encryption.")
-                extra_args = {'ServerSideEncryption': 'AES256'}
-            elif (type(encrypt) is str) and (encrypt.lower() == 'default'):
-                logging.info("Using default bucket encryption.")
-                kms_key_id = self.get_bucket_encryption(bucket_name=bucket)
-                extra_args = kms_key_id
-            elif (type(encrypt) is dict) and encrypt.get('kms_key_id'):
-                logging.info("Using KMS encryption.")
-                extra_args = {
-                    "ServerSideEncryption": encrypt.get('server_side_encryption'),
-                    "SSEKMSKeyId": encrypt.get('kms_key_id')
-                }
-            else:
-                logging.info("No encryption configuration found.")
-                extra_args = {}
+            extra_args = self.set_s3_artifact_encryption(encrypt=encrypt, bucket=bucket)
 
             # Metadata artifact setup
             if metadata:
--- a/gdk/commands/component/PublishCommand.py
+++ b/gdk/commands/component/PublishCommand.py
@@ -49,6 +49,40 @@ class PublishCommand(Command):
             logging.error("Failed to publish new version of the component '{}'".format(self.project_config["component_name"]))
             raise Exception("{}\n{}".format(error_messages.PUBLISH_FAILED, e))
 
+    def get_bucket_encryption(self, bucket_name):
+        """
+        Identifies the default s3 bucket encryption for artifact upload
+
+        Raises an exception when the request is not successful.
+
+        Parameters
+        ----------
+            bucket_name(string): Name of the s3 bucket to pull default encryption from.
+
+        Returns
+        -------
+            encrypt(dict): encryption argument to use with artifact upload
+        """
+        try:
+            encrypt = {}
+            response = self.service_clients["s3_client"].get_bucket_encryption(
+                Bucket=bucket_name
+            )
+            logging.info(f"Getting default encryption for S3 Bucket {bucket_name}.")
+            sse_default = response['ServerSideEncryptionConfiguration']['Rules'][0][
+                'ApplyServerSideEncryptionByDefault']
+            if sse_default.get("SSEAlgorithm") == "AES256":
+                encrypt = {'ServerSideEncryption': sse_default["SSEAlgorithm"]}
+            elif sse_default.get("KMSMasterKeyID"):
+                encrypt = {
+                    "ServerSideEncryption": sse_default["SSEAlgorithm"],
+                    "SSEKMSKeyId": sse_default['KMSMasterKeyID']
+                }
+            return encrypt
+        except Exception as e:
+            logging.error("Failed to get default encryption key for S3 Bucket {}".format(bucket_name))
+            raise Exception("{}\n{}".format(error_messages.PUBLISH_FAILED, e))
+
     def upload_artifacts_s3(self, component_name, component_version):
         """
         Uploads all the artifacts from component artifacts build folder to s3 bucket.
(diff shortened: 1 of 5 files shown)


gdk/commands/component/project_utils.py
```diff
@@ -108,6 +108,8 @@ def get_project_config_values():
     component_build_config = component_config["build"]
     bucket = component_config["publish"]["bucket"]
     region = component_config["publish"]["region"]
+    encrypt = component_config["publish"].get("encrypt")
```
saranyailla, line 111: What are the different possible formats of `encrypt` in the `gdk-config.json`? Could you add that to the description of the PR?

gdk/commands/component/project_utils.py
```diff
@@ -129,6 +131,8 @@ def get_project_config_values():
     vars["component_build_config"] = component_build_config
     vars["bucket"] = bucket
     vars["region"] = region
+    vars["encrypt"] = encrypt
+    vars["metadata"] = metadata
```
saranyailla, line 135: What are the different possible formats of `metadata` in the `gdk-config.json`? IMO, these extra keys should go in publish along with the bucket configuration as these are args applicable to the bucket artifacts. Please provide us with the suggested config file format in the description of the PR.

gdk/commands/component/PublishCommand.py
```diff
@@ -67,6 +101,32 @@ def upload_artifacts_s3(self, component_name, component_version):
...
+                extra_args = {'ServerSideEncryption': 'AES256'}
+            elif (type(encrypt) is str) and (encrypt.lower() == 'default'):
+                logging.info("Using default bucket encryption.")
+                kms_key_id = self.get_bucket_encryption(bucket_name=bucket)
+                extra_args = kms_key_id
+            elif (type(encrypt) is dict) and encrypt.get('kms_key_id'):
+                logging.info("Using KMS encryption.")
+                extra_args = {
+                    "ServerSideEncryption": encrypt.get('server_side_encryption'),
+                    "SSEKMSKeyId": encrypt.get('kms_key_id')
```
saranyailla, line 119: Wondering if kms keys must be passed in the publish command as args or env variables as opposed to having them in the gdk-config file to not expose them accidentally.

gdk/commands/component/PublishCommand.py
```diff
@@ -67,6 +101,32 @@ def upload_artifacts_s3(self, component_name, component_version):
...
+            metadata = self.project_config.get("metadata")
+
+            # Encryption artifact setup
+            if (type(encrypt) is str) and (encrypt.lower() == 'true'):
+                logging.info("Using default encryption.")
+                extra_args = {'ServerSideEncryption': 'AES256'}
+            elif (type(encrypt) is str) and (encrypt.lower() == 'default'):
+                logging.info("Using default bucket encryption.")
+                kms_key_id = self.get_bucket_encryption(bucket_name=bucket)
+                extra_args = kms_key_id
```
saranyailla, line 114: Directly assign bucket encryption to extra_agrs.

gdk/commands/component/PublishCommand.py
```diff
@@ -67,6 +101,32 @@ def upload_artifacts_s3(self, component_name, component_version):
...
+            elif (type(encrypt) is dict) and encrypt.get('kms_key_id'):
+                logging.info("Using KMS encryption.")
+                extra_args = {
+                    "ServerSideEncryption": encrypt.get('server_side_encryption'),
+                    "SSEKMSKeyId": encrypt.get('kms_key_id')
+                }
+            else:
+                logging.info("No encryption configuration found.")
+                extra_args = {}
```
saranyailla, line 123: Move it all the way to the top after L106

gdk/commands/component/PublishCommand.py
```diff
@@ -67,6 +101,32 @@ def upload_artifacts_s3(self, component_name, component_version):
         try:
             bucket = self.project_config["bucket"]
             region = self.project_config["region"]
+            encrypt = self.project_config.get("encrypt")
```
saranyailla, line 104: We can move all of these newly added lines to another helper method.

saranyailla:
PR looks good to me. But to proceed further, you need to make the following changes along with addressing the comments.  

1. Update the commit messages to match the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#examples) format for the GitHub workflow checks to pass. You can also refer to the previous commits of this repository. 

2. Add relevant unit tests and integration tests to maintain the code coverage of the repository so the codecov can pass.

3. Once you're done with 1 and 2, one of the Greengrass engineers will add UATs (mainly test data) for the same.

Let us know if you need more information.

gdk/commands/component/project_utils.py
JustinKuskowski: Updated supported encrypt values in PR description.

gdk/commands/component/project_utils.py
JustinKuskowski: The format would key/value pairs.  

My thought process around this was to keep the bucket / artifact configuration together by putting them into the `gdk-config.json` to keep it close to my code base. Similar to how we do infrastructure as code.

JustinKuskowski: I think it would be beneficial to be able to do it both ways like how the gdk bucket is handled.

gdk/commands/component/PublishCommand.py
JustinKuskowski: I was originally following the process that Serverless Application Model (SAM) encryption method.  Just like the metadata I think you could use both methods.  Configuration based and runtime.

gdk/commands/component/PublishCommand.py
JustinKuskowski: Fixed in new commit.

gdk/commands/component/PublishCommand.py
JustinKuskowski: Fixed in new commit.

gdk/commands/component/PublishCommand.py
JustinKuskowski: Fixed in new commit.

gdk/commands/component/PublishCommand.py
```diff
@@ -49,6 +49,40 @@ def run(self):
...
+            logging.info(f"Getting default encryption for S3 Bucket {bucket_name}.")
+            sse_default = response['ServerSideEncryptionConfiguration']['Rules'][0][
+                'ApplyServerSideEncryptionByDefault']
+            if sse_default.get("SSEAlgorithm") == "AES256":
+                encrypt = {'ServerSideEncryption': sse_default["SSEAlgorithm"]}
+            elif sse_default.get("KMSMasterKeyID"):
+                encrypt = {
+                    "ServerSideEncryption": sse_default["SSEAlgorithm"],
+                    "SSEKMSKeyId": sse_default['KMSMasterKeyID']
+                }
```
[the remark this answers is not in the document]
JustinKuskowski: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-serversideencryptionbydefault.html#cfn-s3-bucket-serversideencryptionbydefault-ssealgorithm

Allowed values: AES256 | aws:kms

gdk/commands/component/PublishCommand.py
JustinKuskowski: Updated encrypt to only use SSE aws:kms.  I removed the other options (i.e. default / true) from the gdk-config.json.

JustinKuskowski:
Just following up on this to make sure no updates are needed on my side.

saranyailla:
[ExtraArgs](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-uploading-files.html#the-extraargs-parameter) parameter in s3 client upload_file operation takes a JSON object as the input. So, instead of adding individual keys from this [ALLOWED_UPLOAD_ARGS](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/customizations/s3.html#boto3.s3.transfer.S3Transfer.ALLOWED_UPLOAD_ARGS) list to the `gdk-config.json` file, the idea is to add a new key called `s3_upload_file_args` under `publish` in gdk config file that takes `ExtraArgs` JSON object directly as `ExtraArgs` is already validated by s3 client `upload_file` operation. 

The following is the output of `gdk component publish` command when used with invalid `s3_upload_file_args` as follows in the `gdk-config.json`

```
      "publish": {
        "bucket": "<BUCKET_NAME>",
        "region": "<AWS_REGION>",
        "s3_upload_file_args": {
          "Meta": {
            "Key": "Value"
          }
        }
      }
```

```
=============================== ERROR ===============================
Could not publish the component due to the following error.
Failed to publish new version of component with the given configuration.
Error while uploading the artifacts to s3 during publish.
Invalid extra_args key 'Meta', must be one of: ACL, CacheControl, ChecksumAlgorithm, ContentDisposition, ContentEncoding, ContentLanguage, ContentType, ExpectedBucketOwner, Expires, GrantFullControl, GrantRead, GrantReadACP, GrantWriteACP, Metadata, ObjectLockLegalHoldStatus, ObjectLockMode, ObjectLockRetainUntilDate, RequestPayer, ServerSideEncryption, StorageClass, SSECustomerAlgorithm, SSECustomerKey, SSECustomerKeyMD5, SSEKMSKeyId, SSEKMSEncryptionContext, Tagging, WebsiteRedirectLocation
```

gdk/commands/component/PublishCommand.py
```diff
@@ -67,6 +67,8 @@ def upload_artifacts_s3(self, component_name, component_version):
         try:
             bucket = self.project_config["bucket"]
             region = self.project_config["region"]
+            s3_upload_file_args = self.project_config.get("s3_upload_file_args")
```
Nelsonochoam, line 70: I just have a comment on the name chosen for the configuration. Currently on the config names there is no reference to `s3`. For instance we do no call `bucket` `s3_bucket`, why start using it for this one?

gdk/static/config_schema.json
```diff
@@ -138,6 +138,10 @@
                                         "us-gov-east-1"
                                     ]
                                 }
+                            },                                
+                            "s3_upload_file_args": {
```
Nelsonochoam, line 142: Even though this brings great flexibility it comes at a cost which is that now we are loosing some type safety, `foo` is not a valid args an we are saying it is now (even though it will have no effect on the request being submitted)

Nelsonochoam: For security is better to only expose allow values for what we want the customer to allow client to do

gdk/commands/component/PublishCommand.py
Nelsonochoam: why not call it `args`? if publish is already in the context of s3

gdk/static/config_schema.json
saranyailla: As mentioned [here](https://github.com/aws-greengrass/aws-greengrass-gdk-cli/pull/99#issuecomment-1430598766), the client operation does this validation. Moreover, we should keep up with S3 supported args and release our `gdk` versions based on that if we couple with it (in schema validation).

gdk/commands/component/PublishCommand.py
saranyailla: This configuration here is very specific to s3 client `upload_file` operation. `args` is too generic in that sense. We want to make it as clear as possible that the args here will be used only during s3 file upload. "publish" is more in the context of creating a new component version where s3 file upload is a sub-task.

saranyailla: Added `file_upload_args` under `options` as discussed offline.

Nelsonochoam (dismissed):

avipinku (approved):

Nelsonochoam (dismissed):

avipinku (approved):

Nelsonochoam (approved):

PR #113, opened 2023-02-13
feat: build options for zip

Nelsonochoam:
**Description of changes:**
Changes the configuration schema to include options to specify excludes for the zip build system. By default the zip build system grabs all the files from the root folder (excluding a few) when the excludes option is specified we will only exclude greengrass-build, recipe and gdk-config files + whatever the customer configures.

Part of functionality being requested by customers


Makefile
```diff
@@ -0,0 +1,5 @@
+lint:
```
Nelsonochoam, line 1: For convenience during local development

gdk/static/config_schema.json
```diff
@@ -64,21 +64,58 @@
                                     "build_system"
                                 ]
                             },
-                            "if": {
-                                "properties": {
-                                    "build_system": {
-                                        "const": "custom"
+                            "allOf": [
```
Nelsonochoam, line 67: This change allow us to have multi conditional if statements which is now required to accept different options per build system type

https://json-schema.org/understanding-json-schema/reference/conditionals.html

tests/gdk/build_system/test_zip_build_configuraiton.py
```diff
@@ -0,0 +1,60 @@
...
+
+        with pytest.raises(Exception):
+            options = {"excludes": []}
+            validate_configuration(self.configuration_base(options))
+
+        with pytest.raises(Exception):
+            options = {}
+            validate_configuration(self.configuration_base(options))
+
+        with pytest.raises(Exception):
```
dgggIoT, line 58: what is the purpose of this sample-like dict?

gdk/static/config_schema.json
```diff
@@ -64,21 +64,58 @@
...
                                 },
-                                "required": [
-                                    "custom_build_command"
-                                ]
-                            }
+                                {
+                                    "if": {
+                                        "properties": {
+                                            "build_system": {
+                                                "const": "custom"
```
dgggIoT, line 105: what build_system will be considered as "custom"? Will "zip" be considerred as "custom" too?

tests/gdk/build_system/test_zip_build_configuraiton.py
Nelsonochoam: That random unsupported value raises an exception

dgggIoT (dismissed):

tests/gdk/build_system/test_zip_build_configuraiton.py
```diff
@@ -0,0 +1,67 @@
...
+                    "build": {
+                        "build_system": "zip"
+                    },
+                    "publish": {
+                        "bucket": "default",
+                        "region": "us-east-1"
+                    },
+                }
+            },
+            "gdk_version": "1.0.0",
```
saranyailla, line 28: Nit: Change this to 1.2.0 though it doesn't really matter.

tests/gdk/build_system/test_zip_build_configuraiton.py
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        if options is not None:
+            base["component"]["com.build.zip"]["build"].setdefault("options", options)
+
+        return base
+
+    def test_valid_configuration_options(self):
```
saranyailla, line 36: Nit: You can use `pytest.mark.parametrize` for passing multiple inputs in these two tests.

gdk/static/config_schema.json
```diff
@@ -64,21 +64,50 @@
...
+                                                "const": "zip"
+                                            }
+                                        }
+                                    },
+                                    "then": {
+                                        "properties": {
+                                            "options": {
+                                                "type": "object",
+                                                "description": "configuraiton options for the zip build system",
+                                                "required": ["exclude"],
```
saranyailla, line 81: I'm not sure how the "options" for "zip" build system are gonna be extended yet. So, I don't think we can go with "required" here for exclude.

Nelsonochoam: Right now it is required given it is the only attribute. Moving forward might not be. This config might change a bit in order for us to support zipping multiple folders

Nelsonochoam: Making a field required to not required is safe (2 way door) if we want to change it later on. The other way around is not.

saranyailla (approved):

tests/gdk/build_system/test_zip_build_configuraiton.py
```diff
@@ -0,0 +1,61 @@
+from pathlib import Path
```
avipinku, line 1: should be `test_zip_build_configuration.py`

gdk/static/config_schema.json
```diff
@@ -64,21 +64,50 @@
...
+                                            "build_system": {
+                                                "const": "zip"
+                                            }
+                                        }
+                                    },
+                                    "then": {
+                                        "properties": {
+                                            "options": {
+                                                "type": "object",
+                                                "description": "configuraiton options for the zip build system",
```
avipinku, line 80: nit: `configuration`

avipinku (approved):

saranyailla (approved):

PR #116, opened 2023-02-14
fix: exit build when the command fails

saranyailla:
**Issue #, if available:**
A part of #97 to exit when the build fails is covered here. 

**Description of changes:**
Exit build when the build command files. This is a bug if a component is published without build. i.e if `gdk component publish` is run `gdk component build` is also run if the component is not built. In this case, if build fails, we publish command should exit as well. Without this change, even if the build fails, component is published.

We can check for the sub-process return code to see if the build succeeded. we're check=True also does the same thing  along with printing the output of the command. 

**Checklist:**

### DIFF
--- a/gdk/commands/component/BuildCommand.py
+++ b/gdk/commands/component/BuildCommand.py
@@ -2,17 +2,18 @@ import json
 import logging
 import platform
 import shutil
-import yaml
 import subprocess as sp
 from pathlib import Path
+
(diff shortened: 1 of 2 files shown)


Nelsonochoam (approved):
This looks good. Are we also going to remove the building part from the publish?

avipinku (approved):

PR #117, opened 2023-02-15
Template command improvements

Nelsonochoam:
**Description of changes:**
Changes the format of how templates are printed. Before
```
1. HelloWorld-python
2. HelloWorld-java
```

After

```
1. HelloWorld (python)
2. HelloWorld (java)
```

The current way the template names are printed leads to confusion causing developers to try to init templates using the `HelloWorld-python` name instead of `HelloWorld` which is the real template name. This tiny usability change will enable a better customer experience by removing a point of confusion.

Unit tests and manually tested


gdk/commands/component/ListCommand.py
```diff
@@ -25,27 +27,36 @@ def run(self):
             return
         raise Exception(error_messages.LIST_WITH_INVALID_ARGS)
 
-    def get_component_list_from_github(self, url):
-        comp_list_response = requests.get(url)
-        if comp_list_response.status_code != 200:
```
dgggIoT, line 30: could you add some explanation here for why removing check of 200?

dgggIoT (approved):

gdk/commands/component/ListCommand.py
Nelsonochoam: That is what raise_for_status() already does

saranyailla (approved):

dgggIoT (approved):

PR #121, opened 2023-02-17
refactor: clean up PublishCommand class

saranyailla:
**Issue #, if available:**

**Description of changes:**
- This change first verifies if the component is built.
- Then updates the project configuration values such as account number, bucket, version in order based on the publish configuration. 

**Checklist:**

### DIFF
--- a/gdk/commands/component/PublishCommand.py
+++ b/gdk/commands/component/PublishCommand.py
@@ -20,34 +20,66 @@ class PublishCommand(Command):
 
     def run(self):
         try:
-            self.project_config["account_number"] = self.get_account_number()
-            if self.arguments["bucket"]:
-                self.project_config["bucket"] = self.arguments["bucket"]
-            else:
-                self.project_config["bucket"] = "{}-{}-{}".format(
-                    self.project_config["bucket"], self.project_config["region"], self.project_config["account_number"]
-                )
-
+            self.is_project_built()
+            self._update_project_config_values()
             component_name = self.project_config["component_name"]
-            component_version = self.get_component_version_from_config()
-            logging.debug(f"Checking if the component '{component_name}' is built.")
-            if not utils.dir_exists(self.project_config["gg_build_component_artifacts_dir"]):
-                logging.warning(
-                    f"The component '{component_name}' is not built.\nSo, building the component before publishing it."
-                )
-                component.build({})
-            logging.info(f"Publishing the component '{component_name}' with the given project configuration.")
-            logging.info("Uploading the component built artifacts to s3 bucket.")
-            self.upload_artifacts_s3(component_name, component_version)
-
-            logging.info(f"Updating the component recipe {component_name}-{component_version}.")
-            self.update_and_create_recipe_file(component_name, component_version)
-
-            logging.info(f"Creating a new greengrass component {component_name}-{component_version}")
-            self.create_gg_component(component_name, component_version)
+            component_version = self.project_config["component_version"]
+            self._publish_component_version(component_name, component_version)
         except Exception as e:
             logging.error("Failed to publish new version of the component '{}'".format(self.project_config["component_name"]))
             raise Exception("{}\n{}".format(error_messages.PUBLISH_FAILED, e))
+
+    def is_project_built(self):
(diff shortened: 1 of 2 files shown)


gdk/commands/component/PublishCommand.py
```diff
@@ -20,35 +20,67 @@ def __init__(self, command_args) -> None:
...
-            logging.info(f"Creating a new greengrass component {component_name}-{component_version}")
-            self.create_gg_component(component_name, component_version)
+            component_version = self.project_config["component_version"]
+            self._publish_component_version(component_name, component_version)
         except Exception as e:
             logging.error("Failed to publish new version of the component '{}'".format(self.project_config["component_name"]))
             raise Exception("{}\n{}".format(error_messages.PUBLISH_FAILED, e))
 
+    def is_project_built(self):
+        # TODO: This method should just warn and proceed. It should not build the component.
```
Nelsonochoam, line 33: Should it continue though? Maybe it should stop instead of trying to publish if not suitable version for deployment is found.

gdk/commands/component/PublishCommand.py
```diff
@@ -20,35 +20,67 @@ def __init__(self, command_args) -> None:
...
         try:
-            self.project_config["account_number"] = self.get_account_number()
-            if self.arguments["bucket"]:
-                self.project_config["bucket"] = self.arguments["bucket"]
-            else:
-                self.project_config["bucket"] = "{}-{}-{}".format(
-                    self.project_config["bucket"], self.project_config["region"], self.project_config["account_number"]
-                )
-
+            self.is_project_built()
```
Nelsonochoam, line 23: Maybe `self.try_build` the `is` prefix makes it seem it returns a boolean

gdk/commands/component/PublishCommand.py
```diff
@@ -20,35 +20,67 @@ def __init__(self, command_args) -> None:
...
+        component_name = self.project_config["component_name"]
+        logging.debug(f"Checking if the component '{component_name}' is built.")
+        if not utils.dir_exists(self.project_config["gg_build_component_artifacts_dir"]):
+            logging.warning(
+                f"The component '{component_name}' is not built.\nSo, building the component before publishing it."
+            )
+            component.build({})
+
+    def _update_project_config_values(self):
+        """'
```
Nelsonochoam, line 43:
```suggestion
        """
```

gdk/commands/component/PublishCommand.py
```diff
@@ -20,35 +20,67 @@ def __init__(self, command_args) -> None:
...
+        Resolve the publish configuration provided in the gdk config file with the CLI arguments passed to the publish command.
+        Arguments provided in the `gdk component publish` commands take precedence over the gdk-config values.
+
+        Parameters
```
Nelsonochoam, line 47: Nit: We should use python typing instead of this return comments, that way code and comments don't go out of sync for return types

gdk/commands/component/PublishCommand.py
```diff
@@ -20,35 +20,67 @@ def __init__(self, command_args) -> None:
...
+        ----------
+            None
+
+        Returns
+        -------
+            None
+        """
+        self._update_account_number()
```
Nelsonochoam, line 55: This is good for now but won't scale well as we add more overridable configuration values. Here the publish command class is being responsible of too many things and breaks single responsibility and open close. We don't have to fix this now but a suggestion to address this is using a [Chain of Responsibility](https://refactoring.guru/design-patterns/chain-of-responsibility#:~:text=Chain%20of%20Responsibility%20is%20a,next%20handler%20in%20the%20chain.)

gdk/commands/component/PublishCommand.py
saranyailla: There are cases when there can be no local artifacts. For eg. 
- All the artifacts in the recipe are already present on s3. (because we first lookup artifact at s3 uri and then locally).
- Component with no artifacts but just a recipe. 

So, in such cases, publish should still continue if there is `greengrass-build/artifacts` directory.

Nelsonochoam (approved):

dgggIoT (approved):

Issue #132, opened 2023-03-01
(gdk): ca-central-1 not supported


gearoid-moore:
**Feature Description**
Making a feature request based off this RePost suggestion:
https://repost.aws/questions/QUe_mJGGF6ScSvyoj0uLixLQ/aws-io-t-pub-sub-sdk-workshop-gdk-tool-fails-in-ca-central-1

I am not the repost thread creator, just someone with the same issue. 
**Use Case**
I used gdk-cli in a publishing pipeline that needs to publish to the canada region.

**Proposed Solution**
As suggested in repost - add ca-central-1 to the schema.

* [ ] :wave: I may be able to implement this feature request
* [ ] :warning: This feature might incur a breaking change

Nelsonochoam:
Hey @gearoid-moore,

Thanks for reaching out and bringing this up. We are working on getting this fixed and will be available on our next patch release 1.2.1 which should be completed in the next couple of days.

Cheers,

Nelson

Nelsonochoam:
@gearoid-moore we just released v1.2.1 which will accept the `ca-central-1` region. Closing this issue

PR #133, opened 2023-03-01
feat: support canada central 1 region

Nelsonochoam:
**Issue #, if available:**
https://github.com/aws-greengrass/aws-greengrass-gdk-cli/issues/132

**Description of changes:**
Remove region enum validation from config if users use an invalid region they will get an error like this one

```
=============================== ERROR ===============================
Could not publish the component due to the following error.
Failed to publish new version of component with the given configuration.
Error while fetching account number from credentials.
Could not connect to the endpoint URL: "https://sts.invalid-region.amazonaws.com/"
```

Users hosting gg on Canada can not use the gdk.

Manually tested publishing a component to ca-central-1


Nelsonochoam:
May

> Do we need this validation at all? Can we just remove it?

> Trying to call Greengrass APIs in a region that we don't exist will fail already, I don't really want us to maintain an additional listing.

True let me see if we can remove that

MikeDombo (approved):

Nelsonochoam:
Reverted removing the validation for regions. @saranyailla had a great point which is that we upload the s3 artifacts before creating the gg component. Not adding the validation will with the current version lead to uploading something to s3 and failing to create the component on a region we don't support.

We can look into improving this later and change it to check if the region is supported by making an api call at the beginning of the whole process.

MikeDombo:
https://stackoverflow.com/questions/56669224/boto3-list-available-services-in-a-region

MikeDombo (approved):

saranyailla (approved):

PR #141, opened 2023-03-03
feat: add case insensitive utils to the recipe files

saranyailla:
**Issue #, if available:**

**Description of changes:**
- Adding a class for`CaseInsensitiveRecipeFile` to read and write JSON and YAML recipe files. 
- `CaseInsensitiveRecipeFile` uses `CaseInsensitiveDict` from `requests` library. 
- By default, `CaseInsensitiveDict`  makes only the top-level key lookup case insensitive. The changes in this PR perform the  case insensitive dict conversion on the nested dictionaries as well.  

This change will allows us to lookup dictionary with case insensitive keys while the original keys are still preserved. 

**Checklist:**


gdk/commands/component/PublishCommand.py
```diff
@@ -255,7 +257,7 @@ def update_and_create_recipe_file(self, component_name, component_version):
         build_recipe = Path(self.project_config["gg_build_recipes_dir"]).joinpath(
             self.project_config["component_recipe_file"].name
         )
-        parsed_component_recipe = project_utils.parse_recipe_file(build_recipe)
+        parsed_component_recipe = self.recipe_factory.get(build_recipe).load()
```
Nelsonochoam, line 260:
```
utils.load_file(path) -> dict

recipe = Recipe.create(dict) -> Recipe
```

.github/workflows/CI.yml
```diff
@@ -52,7 +52,7 @@ jobs:
           pip install -r test-requirements.txt
       - name: Unit Testing CLI
         run: |
-          coverage run --source=gdk -m pytest -v -s tests && coverage xml --fail-under=98
+          coverage run --source=gdk -m pytest -v -s tests && coverage xml --fail-under=90
```
Nelsonochoam, line 55: Why did we lowered this?

gdk/common/recipe_file/CaseInsensitiveRecipe.py
```diff
@@ -0,0 +1,54 @@
...
+                )
+        return recipe
+
+    def __get(self, original: dict) -> CaseInsensitiveDict:
+        __recipe = CaseInsensitiveDict(original)
+        self.__convert_nested_dict(__recipe)
+        return __recipe
+
+    def __get_as_dict(self, cir: CaseInsensitiveDict) -> dict:
+        __recipe = dict(cir)
```
Nelsonochoam, line 38: why the `__` I think there is no need in this case. There is no need to do name mangling here and in most of the other methods. This one should be `recipe` and the names should use single underscore just to indicate they are private. In python using __ doesn't mean you won't be able to access it but that the interpreter would tweak the name when using double `__`

gdk/common/recipe_file/CaseInsensitiveRecipe.py
```diff
@@ -0,0 +1,54 @@
+from requests.structures import CaseInsensitiveDict
+from gdk.common.recipe_file.Recipe import Recipe
+
+
+class CaseInsensitiveRecipe:
```
Nelsonochoam, line 5: Seems weird we are calling this a Recipe given it is not an actual Recipe. According to the OO hierarchy `CaseInsensitiveRecipe` is not a `Recipe` `isinstance(caseInsensitiveRecipe, Recipe` would return false

Maybe an Adapter? dict -> Recipe

gdk/common/recipe_file/CaseInsensitiveRecipe.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def write(self, file_path, content: CaseInsensitiveDict) -> None:
+        """
+        Converts case insensitive dictionary (requests library) to a python dictionary and writes it to given file path.
+        """
+        Recipe().write(file_path, self.__get_as_dict(content))
+
+    def read(self, file_path) -> CaseInsensitiveDict:
+        """
+        Reads contents of the give file as a python dictionary and converts it to a case insensitive
```
Nelsonochoam, line 50:
```suggestion
        Reads contents from the provided file_path as a python dictionary and converts it to a case insensitive
```

gdk/common/recipe_file/CaseInsensitiveRecipe.py
```diff
@@ -0,0 +1,54 @@
...
+        for key, value in recipe.items():
+            if isinstance(value, dict):
+                recipe.update({key: self.__convert_nested_dict(CaseInsensitiveDict(value))})
+            elif isinstance(value, list):
+                recipe.update(
+                    {key: [self.__convert_nested_dict(CaseInsensitiveDict(val)) for val in value if isinstance(val, dict)]}
+                )
+        return recipe
+
+    def __convert_nested_CaseInsensitiveDict(self, recipe: dict) -> dict:
```
Nelsonochoam, line 16:
```suggestion
    def __convert_nested_caseInsensitiveDict(self, recipe: dict) -> dict:
```

gdk/common/recipe_file/CaseInsensitiveRecipe.py
```diff
@@ -0,0 +1,54 @@
+from requests.structures import CaseInsensitiveDict
+from gdk.common.recipe_file.Recipe import Recipe
+
+
+class CaseInsensitiveRecipe:
+    def __convert_nested_dict(self, recipe: CaseInsensitiveDict) -> CaseInsensitiveDict:
```
Nelsonochoam, line 6: Nit: don't call this a recipe given it could be whatever
```suggestion
    def __convert_nested_dict(self, recipe: CaseInsensitiveDict) -> CaseInsensitiveDict:
```

gdk/common/recipe_file/Recipe.py
```diff
@@ -0,0 +1,49 @@
+import json
+import yaml
+from pathlib import Path
+
+
+class Recipe:
+    def write(self, file_path: Path, content: dict) -> None:
```
Nelsonochoam, line 7: This is not very Open Closed this is open to modifications and closed to extensions. I am ok with this though given it is only 2 but we might want to refactor this later on

gdk/common/recipe_file/CaseInsensitiveRecipe.py
```diff
@@ -0,0 +1,54 @@
...
+    def __get_as_dict(self, cir: CaseInsensitiveDict) -> dict:
+        __recipe = dict(cir)
+        self.__convert_nested_CaseInsensitiveDict(__recipe)
+        return __recipe
+
+    def write(self, file_path, content: CaseInsensitiveDict) -> None:
+        """
+        Converts case insensitive dictionary (requests library) to a python dictionary and writes it to given file path.
+        """
+        Recipe().write(file_path, self.__get_as_dict(content))
```
Nelsonochoam, line 46: Would there ever be a case where we don't want a recipe that is case sensitive? I think we should invert the dependencies such that Recipe <- CaseInsensitiveDict  instead of CaseInsensitiveDict <- Recipe.

gdk/common/recipe_file/CaseInsensitiveRecipeFile.py
```diff
@@ -0,0 +1,92 @@
...
+
+    def _write_to_yaml(self, file_path: Path, content: dict) -> None:
+        with open(file_path, "w", encoding="utf-8") as f:
+            yaml.safe_dump(content, f)
+
+    def _is_json(self, file_path: Path) -> bool:
+        return file_path.name.endswith(".json")
+
+    def _is_yaml(self, file_path: Path) -> bool:
+        return file_path.name.endswith(".yaml")
```
MikeDombo, line 56: `.yml` is also valid

gdk/common/recipe_file/CaseInsensitiveRecipeFile.py
```diff
@@ -0,0 +1,92 @@
...
+        """
+        self._write(file_path, self._to_dict(content))
+
+    def read(self, file_path: Path) -> CaseInsensitiveDict:
+        """
+        Reads a JSON or a YAMl file contents as a CaseInsensitiveDict and returns it.
+        """
+        return self._from_dict(self._read(file_path))
+
+    def _write(self, file_path, content):
```
Nelsonochoam, line 20: This method is only called by write, having it as a separate method is not necessary (at least for now)

integration_tests/gdk/common/test_recipe_file/test_CaseInsensitiveRecipeFile.py
```diff
@@ -0,0 +1,63 @@
...
+from unittest import TestCase
+import tempfile
+import pytest
+
+from gdk.common.recipe_file.CaseInsensitiveRecipeFile import CaseInsensitiveRecipeFile
+
+from requests.structures import CaseInsensitiveDict
+
+
+class CaseInsensitiveRecipeFileTest(TestCase):
```
MikeDombo, line 11: validate that when we read in the recipe and write it out, it is unchanged (we don't want all lowercase keys for example)

gdk/common/recipe_file/CaseInsensitiveRecipeFile.py
```diff
@@ -0,0 +1,92 @@
...
+        if self._is_json(file_path):
+            self._write_to_json(file_path, content)
+        elif self._is_yaml(file_path):
+            self._write_to_yaml(file_path, content)
+        else:
+            raise Exception(f"Invalid recipe file : {file_path}. Recipe file must be in json or yaml format.")
+
+    def _read(self, file_path):
```
Nelsonochoam, line 28: We could simplify this class a bit to reduce the cyclomatic complexity and the amount of indirection while making it a bit more declarative by doing something along this lines

```
read = {
   ".json":  json.loads,
   ".yaml":  yaml.safe_load,
  ".yml":  yaml.safe_load,
}

write = {
    ".yaml": yaml.safe_dump,
    ".yml":  yaml.safe_dump,
    ".json":  json.dump
}

def read(file_path: Path) -> CaseInsensitiveDict:
    _read = read.get(file_path.suffix)

    if _read is None:
          raise Exception(f"Invalid recipe file : {file_path}. Recipe file must be in json or yaml format.")

            return self._from_dict(_read(f.read()))


def write(content: file_path: CaseInsensitiveDict) -> None:
    _write = write.get(file_path.suffix)

    if _write in None:
           raise Exception(f"Invalid recipe file : {file_path}. Recipe file must be in json or yaml format.")

     with open(file_path, "w", encoding="utf-8") as f:
            _write(content, f)
```

integration_tests/gdk/common/test_recipe_file/test_CaseInsensitiveRecipeFile.py
saranyailla: Added a test that compares the file contents of original yaml and updated yaml.

gdk/common/recipe_file/CaseInsensitiveRecipeFile.py
saranyailla: Agreed. Without `_write()` and `_read()`, we'll have to convert to_dict or from_dict for both json and yaml operations. Just doing that only one time with this additional _write and _read.

MikeDombo (approved):

gdk/common/recipe_file/CaseInsensitiveRecipeFile.py
saranyailla: We're using some additional arguments for both `json.dump` and `yaml.safe_dump`. They also take the file argument in different order.

MikeDombo (approved):

gdk/common/CaseInsensitive.py
```diff
@@ -0,0 +1,98 @@
...
+                            for val in value
+                            if isinstance(val, CaseInsensitiveDict)
+                        ]
+                    }
+                )
+        return dictObj
+
+
+class CaseInsensitiveRecipeFile:
+    def write(self, file_path: Path, content: CaseInsensitiveDict) -> None:
```
Nelsonochoam, line 50: I am ok with this given we only support 2 formats but if we ever get to more than 2 and support something like `toml` or `xml` this needs to be refactored.

Nelsonochoam (approved):

MikeDombo (approved):

Issue #138, opened 2023-03-02
(gdk): S3 bucket creation fails if account is on bucket limit


gearoid-moore:
**Describe the bug**
It seems like the s3 bucket creation exception handling needs tweaking, exception catch for bucket limit supersedes situation where bucket exists and at limit. 

**To Reproduce**
Create an account with the limit of S3 buckets. Try to publish a component to a bucket that already exists ( meaning no new buckets should be created ) . It will fail with an error that it cannot create a bucket. 

**Expected behavior**
Using the full bucket quota is a valid behaviour and should be supported. 

**Actual behavior**
Component does not publish . 

**Environment**
 - OS:  Ubuntu 20.04
 - Nucleus version: 2.9.3

**Additional context**
Cannot publish components using gdk as I am maxed out on buckets.

saranyailla:
This is fixed in GDK v1.2.2.

PR #146, opened 2023-03-14
fix: check for bucket's existence

saranyailla:
**Issue #, if available:**
#138 

**Description of changes:**
Base commit https://github.com/aws-greengrass/aws-greengrass-gdk-cli/commit/0d28f97566d648f0dede47903e61b9cef919b7f9 from PR https://github.com/aws-greengrass/aws-greengrass-gdk-cli/pull/124

Since Greengrass supports only artifacts from the same account and region as of the components', before creating the bucket, GDK now checks for the bucket location using `get_bucket_location` API. This API works only when it is called by the bucket owner. The output of this API contains location (region) of the bucket. 

- If the user of the command is not the owner of the bucket, `get_bucket_location` throws an exception
  - `404 - Not found` Bucket does not exist, GDK can create a new bucket.
  - `403 - Forbidden` Bucket already exists but is owned by someone. So, GDK exits as user should provide a new bucket name. 
- If the user of the command is the owner of the bucket, GDK will further verify the location of the bucket. 
  - If the region obtained using this `get_bucket_location` API is same as that of the region in GDK configuration, GDK will use the bucket without trying to create a new one. 
  - Otherwise, it means that the bucket is owned by the user but exists in a different region (Not supported by GG), so GDK exits as user should provide a new bucket name. 

So, with this change, GDK will create a new bucket only when it does not exist in any region or any account.

This change is needed for the customers to use existing bucket without having to create one. 
Without this change, even when the bucket exists and is provided in the GDK configuration, customers may not be able to publish their artifacts to the bucket as GDK may fail to create the bucket first. 

Bucket creation during component creation may fail due to several reasons such as (even when a valid bucket name is provided in the configuration)  :
- Account reached max bucket limit. 
- User may not have permission to create a new bucket.

**Checklist:**

### DIFF
--- a/gdk/aws_clients/S3Client.py
+++ b/gdk/aws_clients/S3Client.py
@@ -27,46 +27,20 @@ class S3Client:
         -------
             None
         """
+        if self.valid_bucket_for_artifacts_exists(bucket, region):
+            logging.info("Not creating an artifacts bucket as it already exists.")
+            return
+
         try:
             if region is None or region == "us-east-1":
                 self.s3_client.create_bucket(Bucket=bucket)
             else:
                 location = {"LocationConstraint": region}
                 self.s3_client.create_bucket(Bucket=bucket, CreateBucketConfiguration=location)
-        except ClientError as exc:
-            if exc.response["Error"]["Code"] == "BucketAlreadyExists":
-                logging.error("Bucket already exists. Please provide a different name for the bucket in the configuration.")
-            elif exc.response["Error"]["Code"] == "BucketAlreadyOwnedByYou":
-                if self.bucket_exists_in_same_region(bucket, region):
-                    logging.info("Not creating an artifacts bucket as it already exists.")
-                    return
-                logging.error("Cannot create the artifacts bucket '%s' as it is already owned by you in other region.", bucket)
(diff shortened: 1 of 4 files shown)


gdk/aws_clients/S3Client.py
```diff
@@ -27,47 +27,21 @@ def create_bucket(self, bucket, region):
...
-                logging.error("Bucket already exists. Please provide a different name for the bucket in the configuration.")
-            elif exc.response["Error"]["Code"] == "BucketAlreadyOwnedByYou":
-                if self.bucket_exists_in_same_region(bucket, region):
-                    logging.info("Not creating an artifacts bucket as it already exists.")
-                    return
-                logging.error("Cannot create the artifacts bucket '%s' as it is already owned by you in other region.", bucket)
-            raise Exception(exc) from exc
         except Exception as exc:
-            print(exc)
             logging.error("Failed to create the bucket '%s' in region '%s'", bucket, region)
```
MikeDombo, line 41: use logging.exception so that it will print the actual exception. Right now we have no idea what went wrong.

Nelsonochoam (approved):

MikeDombo (approved):

PR #151, opened 2023-04-05
refactor: update & create case-insensitive recipes

saranyailla:
**Issue #, if available:**

**Description of changes:**
- Create recipe transformer classes to delegate the responsibilities of Build and Publish classes. 
- `BuildRecipeTransformer` and `PublishRecipeTransformer` classes update the recipe properties such as component name, version, artifact URIs based on the command used. Recipe properties are looked up case-insensitively. When a  recipe property is updated, it's original case is preserved. 

**Checklist:**

### DIFF
--- a/gdk/commands/component/BuildCommand.py
+++ b/gdk/commands/component/BuildCommand.py
@@ -11,7 +11,7 @@ import gdk.common.utils as utils
 from gdk.build_system.BuildSystem import BuildSystem
 from gdk.build_system.Zip import Zip
 from gdk.commands.Command import Command
-from gdk.commands.component.recipe_generator.BuildRecipeGenerator import BuildRecipeGenerator
+from gdk.commands.component.transformer.BuildRecipeTransformer import BuildRecipeTransformer
 
 
 class BuildCommand(Command):
@@ -19,7 +19,7 @@ class BuildCommand(Command):
         super().__init__(command_args, "build")
         self.project_config = project_utils.get_project_config_values()
         self.supported_build_sytems = project_utils.get_supported_component_builds()
-        self.build_recipe_generator = BuildRecipeGenerator(self.project_config)
+        self.build_recipe_transformer = BuildRecipeTransformer(self.project_config)
 
     def run(self):
         """
@@ -102,7 +102,7 @@ class BuildCommand(Command):
         try:
             # Build the project with specified build system
             self.run_build_command()
-            self.build_recipe_generator.generate(self._get_build_folder_by_build_system())
+            self.build_recipe_transformer.transform(self._get_build_folder_by_build_system())
         except Exception as e:
             raise Exception("""{}\n{}""".format(error_messages.BUILD_FAILED, e))
 
--- a/gdk/commands/component/BuildCommand.py
+++ b/gdk/commands/component/BuildCommand.py
@@ -1,11 +1,8 @@
-import json
 import logging
 import platform
-import shutil
 import subprocess as sp
 from pathlib import Path
 
-import yaml
 
 import gdk.commands.component.project_utils as project_utils
 import gdk.common.consts as consts
@@ -14,6 +11,7 @@ import gdk.common.utils as utils
 from gdk.build_system.BuildSystem import BuildSystem
 from gdk.build_system.Zip import Zip
 from gdk.commands.Command import Command
+from gdk.commands.component.recipe_generator.BuildRecipeGenerator import BuildRecipeGenerator
 
 
 class BuildCommand(Command):
@@ -21,6 +19,7 @@ class BuildCommand(Command):
         super().__init__(command_args, "build")
         self.project_config = project_utils.get_project_config_values()
         self.supported_build_sytems = project_utils.get_supported_component_builds()
+        self.build_recipe_generator = BuildRecipeGenerator(self.project_config)
 
     def run(self):
         """
@@ -103,12 +102,7 @@ class BuildCommand(Command):
         try:
             # Build the project with specified build system
             self.run_build_command()
-
-            # From the recipe, copy necessary artifacts (depends on build system) to the build folder .
-            self.find_artifacts_and_update_uri()
-
-            # Update recipe file with component configuration from project config file.
-            self.create_build_recipe_file()
+            self.build_recipe_generator.generate(self._get_build_folder_by_build_system())
         except Exception as e:
             raise Exception("""{}\n{}""".format(error_messages.BUILD_FAILED, e))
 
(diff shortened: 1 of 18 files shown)


gdk/commands/component/recipe_generator/BuildRecipeGenerator.py
```diff
@@ -0,0 +1,137 @@
...
+import shutil
+from pathlib import Path
+from gdk.common.CaseInsensitive import CaseInsensitiveRecipeFile, CaseInsensitiveDict
+
+import gdk.commands.component.project_utils as project_utils
+import gdk.common.consts as consts
+import gdk.common.utils as utils
+
+
+class BuildRecipeGenerator:
```
Nelsonochoam, line 11: I think we are using the word "Recipe" too generously. Calling generate which I assume would be a factory method does not actually return anything it just mutates whatever is read from the `component_recipe_file`. There is no encapsulation here and we are mostly mutating  data.

We should be looking a some creational patterns  to organize this the code for example using builder aggregation like this [example](https://refactoring.guru/design-patterns/builder#:~:text=with%20the%20director.-,pseudocode,-This%20example%20of)

Nelsonochoam: Also if you think about it. In the context of the GDK, the user is not providing us a recipe the provide us with a RecipeTemplate which we then transform into a template. So maybe CaseInsensitiveRecipeFile is not a good name

gdk/commands/component/recipe_generator/BuildRecipeGenerator.py
```diff
@@ -0,0 +1,137 @@
...
+    def __init__(self, project_config) -> None:
+        self.project_config = project_config
+
+    def generate(self, build_folders):
+        component_recipe = CaseInsensitiveRecipeFile().read(self.project_config["component_recipe_file"])
```
Nelsonochoam, line 16: Seems like what we are calling generators is continuously reading and writing values to the same file, ideally we read once, make the all the transformations and then write. 

Read recipe template -> transformation pipeline -> Recipe generated -> Write recipe file

gdk/commands/component/recipe_generator/BuildRecipeGenerator.py
```diff
@@ -0,0 +1,137 @@
...
+        Build command fails when the artifacts are neither not found in local both folders not on s3.
+        """
+        logging.info("Copying over the build artifacts to the greengrass component artifacts build folder.")
+        logging.info("Updating artifact URIs in the recipe.")
+        if "Manifests" not in parsed_component_recipe:
+            logging.debug("No 'Manifests' key in the recipe.")
+            return
+        manifests = parsed_component_recipe["Manifests"]
+        s3_client = None
+        for manifest in manifests:
```
Nelsonochoam, line 47: All this nested conditionals here are a code smell. The `parsed_component_recipe` is a misleading, when we parse data we should be converting the data from one format to a format we know is already valid, otherwise passing a CaseInsensitiveDict dict around forces the received to always have to do validations 

Here is an article that explains the mantra of parse, not validate https://towardsdatascience.com/parse-dont-validate-f559372cca45

Nelsonochoam: If you seen the code below the other "Generator" is having to do similar validation all over again because we are not properly parsing and we are relying to much on the dictionary.

gdk/commands/component/recipe_generator/BuildRecipeGenerator.py
Nelsonochoam: PS: If we don't want to do this changes now, I would suggest not creating this classes and just making the changes on the methods they were before, later on we could refactor it

saranyailla: These classes are created only to handle the responsibility of updating and create the recipe and simplify testing. 
The generator classes create a recipe file in the greengrass-build folders after updating the recipe with the project configuration.

saranyailla: I'm not introducing new interfaces or design patterns by only moving the code around for testing.

gdk/commands/component/recipe_generator/BuildRecipeGenerator.py
saranyailla: `generate` is only used once. It does what you mentioned -  reads, updates everything that is needed and then write once.

integration_tests/gdk/components/test_integ_BuildCommand.py
```diff
@@ -1,7 +1,7 @@
 import json
 from pathlib import Path
 from shutil import Error
-from unittest.mock import mock_open, patch
+from gdk.commands.component.transformer.BuildRecipeTransformer import BuildRecipeTransformer
```
Nelsonochoam, line 4: If you re export on the __init__.py you can avoid retyping the names

```python
#__init__.py
from .BuildRecipeTransformer import BuildRecipeTransformer
```

Then you imports here could be `from gdk.commands.component.transformer import BuildRecipeTransformer`

gdk/commands/component/transformer/BuildRecipeTransformer.py
```diff
@@ -0,0 +1,137 @@
+import logging
+import shutil
+from pathlib import Path
+from gdk.common.CaseInsensitive import CaseInsensitiveRecipeFile, CaseInsensitiveDict
```
Nelsonochoam, line 4: We should start using re exports from `__init__.py` files. On the `gdk.commont.___init__.py` file do

```python
from . CaseInsensitive import *
```

then you could 

```python
from gdk.common import CaseInsensitiveRecipeFile, CaseInsensitiveDict
```

gdk/commands/component/BuildCommand.py
```diff
@@ -14,13 +11,15 @@
...
 from gdk.commands.Command import Command
+from gdk.commands.component.transformer.BuildRecipeTransformer import BuildRecipeTransformer
 
 
 class BuildCommand(Command):
     def __init__(self, command_args) -> None:
         super().__init__(command_args, "build")
         self.project_config = project_utils.get_project_config_values()
         self.supported_build_sytems = project_utils.get_supported_component_builds()
+        self.build_recipe_transformer = BuildRecipeTransformer(self.project_config)
```
Nelsonochoam, line 22: I understand passing the entire project config makes it easy but it opens the door for internally coupling to the self.project_config structure. Could we just pass it what it needs?

gdk/commands/component/PublishCommand.py
```diff
@@ -119,14 +118,24 @@ def _update_region(self):
...
     def _update_component_version(self):
         self.project_config["component_version"] = self.get_component_version_from_config()
+        self._update_publish_recipe_file_path()
+
+    def _update_publish_recipe_file_path(self):
+        ext = self.project_config["component_recipe_file"].name.split(".")[-1]
+        publish_recipe_file_name = "{}-{}.{}".format(
+            self.project_config["component_name"], self.project_config["component_version"], ext
+        )  # Eg. HelloWorld-1.0.0.yaml
+        self.project_config["publish_recipe_file"] = (
```
Nelsonochoam, line 128: Is there a way we could not mutate the project_config? Mutating this is specially bad given we treat this guy as a god object that gets passed around. If not lets add a  TODO, to avoid doing this

gdk/commands/component/transformer/BuildRecipeTransformer.py
```diff
@@ -0,0 +1,137 @@
...
+        """
+        logging.info("Copying over the build artifacts to the greengrass component artifacts build folder.")
+        logging.info("Updating artifact URIs in the recipe.")
+        if "Manifests" not in parsed_component_recipe:
+            logging.debug("No 'Manifests' key in the recipe.")
+            return
+        manifests = parsed_component_recipe["Manifests"]
+        s3_client = None
+        for manifest in manifests:
+            if "Artifacts" not in manifest:
```
Nelsonochoam, line 48: We already chatted about ways for improve this and mentioned it is ok for now given we are just moving things around

gdk/commands/component/BuildCommand.py
saranyailla: BuildReciperTransformer needs 9 properties of `self.project_config`. So, I think we can pass the full object for now

gdk/commands/component/transformer/BuildRecipeTransformer.py
saranyailla: Yeah, will make this change in the following PRs.

MikeDombo (approved):

gdk/commands/component/BuildCommand.py
Nelsonochoam: We should be able to pass less as we split this in the future. It is ok for now

Nelsonochoam (approved):

PR #159, opened 2023-04-10
fix: bucket creation and parsing lists in recipe

uhinze:
**Issue #, if available:** None

**Description of changes:** Fix doc and implementation for non-existing buckets. Fix list expansion.

**Why is this change necessary:** New projects without bucket don't work

**How was this change tested:** Manual test with non-existing and non-accessible bucket, unit tests changed

Issue 1 reproduction:
- Run `component build` and `component publish` on a new project without an existing bucket

Issue 2 reproduction:
- Create a component recipe like this: https://github.com/awslabs/aws-greengrass-labs-bluetooth-gateway/blob/main/src/recipe.json
- Run `component build`
- Check `greengrass-build/recipe.json` and see that `operations` and `resources` are now empty lists
--> this will cause the component deployment to fail

**Checklist:**
- [x] Updated the README if applicable

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Configure AWS CLI with your credentials as shown here - https://docs.aws.amazon.
         "gdk_version": "1.0.0"
     }
    ```
-   2. Replace `<PLACEHOLDER_AUTHOR>` with your name, `<PLACEHOLDER_BUCKET>` with a s3 bucket name and `<PLACEHOLDER_REGION>` with an aws region.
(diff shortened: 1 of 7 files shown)


jfduque:
Can confirm both issues. 

Can also confirm that this PR fixes both issues. Looks like they were introduced in 1.2.2

saranyailla:
PR looks good to me. But to proceed further, please update the commit messages to match the [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#examples) format for the GitHub workflow checks to pass. You can also refer to the previous commits of this repository.

Nelsonochoam (approved):

PR #164, opened 2023-04-21
feat: emit basic install metric

Nelsonochoam:
## Description
Emits install metric by default telemetry is disabled. There will be follow up requests  to ensure we don't emit metric on the same thread and that if emitting fails it doesn't stop the command from running.

## How was this tested?
unit + integ tests

### DIFF
--- a/gdk/CLIParser.py
+++ b/gdk/CLIParser.py
@@ -5,6 +5,7 @@ import gdk.common.consts as consts
 import gdk.common.model_actions as model_actions
 import gdk.common.parse_args_actions as parse_args_actions
 import gdk.common.utils as utils
+from gdk.telemetry.emit import Emit
 
 
 class ArgumentParser(argparse.ArgumentParser):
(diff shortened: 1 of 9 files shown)


gdk/CLIParser.py
```diff
@@ -180,6 +181,8 @@ def _add_common_args_for_all_commands(self):
 
 def main():
     try:
+        Emit().installed_metric()
```
MikeDombo, line 184: why "installed"? this just fires every time you do anything at all right because this is in main.

MikeDombo (commented):
Suggest "execution" instead of "installed", this metric does not track installs, it tracks how many times GDK is invoked/executed.

gdk/CLIParser.py
Nelsonochoam: I mentioned it on the readme. I know this would fire every time. Next PR will address that. Right now it is disabled. This is similar to what SAM_CLI does they do it on their main and record that they emitted it and after that they stop.

MikeDombo (approved):

saranyailla (approved):

PR #165, opened 2023-04-24
Emit install metric only once.

Nelsonochoam:
**Description of changes:**
This PR adds Runtime Config which is basically a singleton we can use to store information we need to store for subsequent runs. Currently the object is not thread safe but we could implement locking later.  It is used so we record if we have already emitted the installed metrics and will be used to record if we have shown the user the install prompt.

The runtime config is just meant to hold key value pairs. If we ever decided to change keys in the config in a future version we would have to nix it and recreate it or use a different name and ignore the old one. The only harm tampering with the file is that the user MAY cause it to trigger other an install metric one more time. When we load it we only take in the keys that we define on the app, the rest are ignored

* Ensure there are certain actions we just perform once across multiple runs of the CLI.

Unit and integration tests

### DIFF
--- a/gdk/runtime_config.py
+++ b/gdk/runtime_config.py
@@ -1,0 +1,91 @@
+import json
+import logging
+import os
+from enum import Enum
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigKey(Enum):
+    INSTALLED = 'installed'
+
+
+class Singleton(type):
+    """
+    Metaclass makes a class a singleton.
+    """
+
+    def __init__(cls, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        cls.__instance = None
+
+    def __call__(cls, *args, **kwargs):
+        """
+        Called every time a new instance of a class is being implemented.
+        """
+        force_create = kwargs.get('force_create', False)
+
+        if not cls.__instance or force_create is True:
+            cls.__instance = super().__call__(*args, **kwargs)
+
+        return cls.__instance
+
+
+class RuntimeConfig(metaclass=Singleton):
+    """
+    Wrapper around a dict that persists values used by the application by flushing them to
+    disk.
+    """
+    DEFAULT_CONFIG_FILE_NAME = 'runtime.json'
+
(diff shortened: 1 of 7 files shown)


gdk/telemetry/emit.py
```diff
@@ -9,8 +11,16 @@ class Emit:
...
 
     def __init__(self, emiter: ITelemetry = None):
+        self.runtime_config = RuntimeConfig()
         self._emiter = emiter or Telemetry()
 
     def installed_metric(self):
+        """
+        Sends an installed metric only once after the cli has been installed
+        """
+        if self.runtime_config.has(ConfigKey.INSTALLED):
```
MikeDombo, line 21: check for this version specifically?

Nelsonochoam: Yeah I want to write a test on the next PR so we emit it if the version has changed

Nelsonochoam: Actually I will just add it one this one

MikeDombo (approved):

gdk/runtime_config.py
```diff
@@ -0,0 +1,102 @@
...
+import logging
+import os
+from enum import Enum
+from pathlib import Path
+
+logger = logging.getLogger(__name__)
+
+
+class ConfigKey(Enum):
+    INSTALLED = 'installed'
```
saranyailla, line 11: `installed_gdk_version`

gdk/runtime_config.py
```diff
@@ -0,0 +1,102 @@
...
+
+GDK_RUNTIME_CONFIG_DIR = "__GDK_RUNTIME_CONFIG_PATH"
+
+
+class RuntimeConfig(metaclass=Singleton):
+    """
+    Wrapper around a dict that persists values used by the application by flushing them to
+    disk.
+    """
+    DEFAULT_CONFIG_FILE_NAME = 'runtime.json'
```
saranyailla, line 43: If this is only telemetry related, we can give it some other name like `telemtry.json`. If not, `runtime.json` works.

saranyailla (approved):

PR #167, opened 2023-04-25
Implement sigv4 signed requests to api gateway on the telemetry adapter

Nelsonochoam:
**Description of changes:**
This PR changes how we make requests to the telemetry endpoint so that every single request that is made needs to be sigv4 signed. On the tests the test telemetry server does some validation for sigv4 which the real service would do once deployed.

Ensure only AWS users can perform requests to the gdk telemetry service endpoint.

Integration tests


MikeDombo (approved):
LGTM

saranyailla (approved):

PR #186, opened 2023-05-18
feat: identify testing jar from the build folder

saranyailla:
**Issue #, if available:**

**Description of changes:**
This change helps gdk identify testing jar from the build output. The testing module build process may generate multiple jar files based on the customer's configuration. When customer runs `gdk test run` command, gdk should identify the testing jar and run it for the customers with the provided options. 

gdk first checks for `uat-features-1.0.0.jar ` in the build folder as that's the default output if the module is built from the sample template. 

If the default jar is not present, then all the `.jar` files in the build folders are run with `--help` command and their output is tested.If the output contains `gg-test`, then it is identified as a testing jar and the tests are run with it. 

**Checklist:**

### DIFF
--- a/gdk/commands/test/RunCommand.py
+++ b/gdk/commands/test/RunCommand.py
@@ -35,7 +35,7 @@ class RunCommand(Command):
             logging.info("Downloading latest nucleus archive from url %s", self._nucleus_archive_link)
             URLDownloader(self._nucleus_archive_link).download(_nucleus_path)
 
-        self._run_testing_jar()
+        self.run_testing_jar()
 
     def _is_test_module_built(self) -> bool:
         """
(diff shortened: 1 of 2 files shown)


urvashijain18 (approved):

gdk/commands/test/RunCommand.py
```diff
@@ -57,12 +57,12 @@ def _should_download_nucleus_archive(self, _nucleus_path: Path) -> bool:
...
         return _nucleus_path == Path(self._config.default_nucleus_archive_path).resolve() and not _nucleus_path.exists()
 
-    def _run_testing_jar(self) -> None:
+    def run_testing_jar(self) -> None:
         """
         Run the testing jar from the build folder using the configured options.
         """
-        # TODO: identify jar from the build output
-        _jar_path = str(self._test_build_directory().joinpath("uat-features-1.0.0.jar").resolve())
+        _jar_path = self._identify_testing_jar().__str__()
```
jcosentino11, line 64:
```suggestion
        _jar_path = self._identify_testing_jar().relative_to(Path.cwd())
```
not sure how reliable string representation is

jcosentino11:
```suggestion
        _jar_path = str(self._identify_testing_jar())
```
nit:

jcosentino11 (approved):

PR #190, opened 2023-05-22
fix: do not validate region in schema  & use ssm client to get supported regions

saranyailla:
**Issue #, if available:**

**Description of changes:**
- Remove region validation from schema.
- Get greengrass supported regions using ssm client and validate it in the publish command. 

**Checklist:**

### DIFF
--- a/gdk/aws_clients/AWSClient.py
+++ b/gdk/aws_clients/AWSClient.py
@@ -1,0 +1,18 @@
+import boto3
+
+service_clients = {}
+
+
+def ssm(region_name="us-east-1"):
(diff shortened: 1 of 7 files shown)


jcosentino11 (dismissed):

gdk/aws_clients/AWSClient.py
```diff
@@ -0,0 +1,18 @@
+import boto3
+
+service_clients = {}
+
+
+def ssm(region_name="us-east-1"):
+    return _get_client("ssm", region_name)
```
MikeDombo, line 7: why ssm?

PR #200, opened 2023-06-27
refactor: refactor client methods

saranyailla:
**Issue #, if available:**

**Description of changes:**
- Refactor AWS client methods in such a way that they don't depend on project configuration. 
- Update tests for the same. 

**Checklist:**

### DIFF
--- a/gdk/aws_clients/Greengrassv2Client.py
+++ b/gdk/aws_clients/Greengrassv2Client.py
@@ -10,7 +10,7 @@ class Greengrassv2Client:
     def __init__(self, _region):
         self.client = boto3.client("greengrassv2", region_name=_region)
 
-    def get_highest_component_version_(self, component_arn) -> str:
+    def get_highest_cloud_component_version(self, component_arn) -> str:
         """
         Gets highest version of the component from the sorted order of its versions from an account in a region.
 
--- a/gdk/aws_clients/Greengrassv2Client.py
+++ b/gdk/aws_clients/Greengrassv2Client.py
@@ -1,4 +1,5 @@
 import logging
+import boto3
 
 
 class Greengrassv2Client:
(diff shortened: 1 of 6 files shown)


gdk/aws_clients/Greengrassv2Client.py
```diff
@@ -1,65 +1,49 @@
...
     Greengrasv2 client utils wrapper
     """
 
-    def __init__(self, project_configuration, service_clients):
-        self.project_config = project_configuration
-        self.greengrass_client = service_clients["greengrass_client"]
+    def __init__(self, _region):
+        self.client = boto3.client("greengrassv2", region_name=_region)
 
-    def get_highest_component_version_(self) -> str:
```
MikeDombo, line 13: maybe rename this since it is doing something different than before. Name `get_highest_cloud_component_version`?

gdk/commands/component/PublishCommand.py
```diff
@@ -175,7 +191,11 @@ def get_next_version(self):
         logging.debug("Fetching private components from the account.")
         try:
             c_name = self.project_config["component_name"]
-            c_next_patch_version = self.greengrass_client.get_highest_component_version_()
+            region = self.project_config["region"]
+            account_num = self.project_config["account_number"]
+            component_arn = f"arn:aws:greengrass:{region}:{account_num}:components:{c_name}"
```
MikeDombo, line 196: partitions.

gdk/commands/component/PublishCommand.py
```diff
@@ -136,27 +136,43 @@ def _publish_component_version(self, component_name, component_version):
...
+            logging.info("No artifacts found in the component build folder. Skipping the artifact upload step")
+            return
+
         logging.info(
-            f"Uploading component artifacts to S3 bucket: {bucket}. If this is your first time using this bucket, add the"
-            " 's3:GetObject' permission to each core device's token exchange role to allow it to download the component"
-            f" artifacts. For more information, see {utils.doc_link_device_role}."
+            (
+                "Uploading component artifacts to S3 bucket: %s. If this is your first time using this bucket, add the"
+                " 's3:GetObject' permission to each core device's token exchange role to allow it to download the component"
```
parrottsquawk, line 158: Is there any way to automate this?

MikeDombo: No, we don't know what role they're setup to use, and we don't want to blindly add wide permissions either.

gdk/commands/component/PublishCommand.py
```diff
@@ -136,27 +136,43 @@ def _publish_component_version(self, component_name, component_version):
...
+        component_name = self.project_config["component_name"]
+        component_version = self.project_config["component_version"]
+        options = self.project_config["options"]
+        s3_upload_file_args = options.get("file_upload_args", dict())
 
-        if len(build_component_artifacts) != 0:
-            self.s3_client.create_bucket(bucket, region)
-            self.s3_client.upload_artifacts(build_component_artifacts)
+        for artifact in build_component_artifacts:
+            s3_file_path = f"{component_name}/{component_version}/{artifact.name}"
```
parrottsquawk, line 173: should we be validating the `s3_file_path` is correctly formed e.g. customer does not have an artifact name of `my_component/`

parrottsquawk (dismissed):

gdk/commands/component/PublishCommand.py
parrottsquawk: Cool, it's a hard thing to remember and find. I'm glad we have a comment here at least. :)

gdk/commands/component/PublishCommand.py
parrottsquawk: Can you elaborate? :)

MikeDombo: Saranya knows :)

`aws` is only valid for most regions, but not all. cn-north-1 for example needs `aws-cn` instead of `aws`.

parrottsquawk: Ahhh, I see. Thanks for the info 👍

saranyailla: [More info on ARNs.](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference-arns.html#:~:text=type%3Aresource-id-,partition,-The%20partition%20in).

MikeDombo (approved):

gdk/commands/component/PublishCommand.py
saranyailla: Here, GDK constructs the s3 file path only when the artifact exists. All the config values provided by the customer are validated at the very beginning of the command execution and component arn is constructed only after that.

urvashijain18 (approved):

PR #206, opened 2023-07-17
fix: validate component version before uploading the artifacts

saranyailla:
**Issue #, if available:**

**Description of changes:**
If a version other than `NEXT_PATCH` is provided in the config, it is never validated until the end of the command execution. i.e if the component version provided in the config already exists, built artifacts are uploaded to s3, recipe is created and then the component creation fails as the version already exists. 

This change validates if the component version already exists by using `get_component` API at the beginning of the `gdk component publish` command execution. 

This change requires the customers to add a new permission `GetComponent` to their existing role. 

**Checklist:**

### DIFF
--- a/gdk/aws_clients/Greengrassv2Client.py
+++ b/gdk/aws_clients/Greengrassv2Client.py
@@ -1,5 +1,6 @@
 import logging
 import boto3
+from botocore.exceptions import ClientError
 
 
 class Greengrassv2Client:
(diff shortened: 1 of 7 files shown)


MikeDombo:
Do we still want this change?

saranyailla:
With this change, customers need to add an extra permission to their role or the command fails. If we don't want to break like that, GDK CLI should attempt to validate component version before trying to create one to fail fast. If that validation fails due to permission issues, it should fallback to previous way which is validating the component version at the time of creation. However, I don't think we should complicate this simple component version creation flow for this. 

We can close this for now and improve this by asking customers to add an extra permission when we do some major version upgrades.

PR #214, opened 2023-07-19
fix: validate OTF version before downloading the template

saranyailla:
**Issue #, if available:**

**Description of changes:**
- Validate the OTF version provided in the config and the command arguments using `semver` library before downloading the template from GitHub. 
- Update third party licenses.
- Add unit tests for the same. 

**Checklist:**

### DIFF
--- a/THIRD-PARTY-LICENSES
+++ b/THIRD-PARTY-LICENSES
@@ -1,0 +1,31 @@
+
+** semver3;  -- https://semver.org 
+©2023, Kostiantyn Rybnikov and all.
+
+Copyright (c) 2013, Konstantine Rybnikov
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
+
+  Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
+  Redistributions in binary form must reproduce the above copyright notice, this
+  list of conditions and the following disclaimer in the documentation and/or
+  other materials provided with the distribution.
+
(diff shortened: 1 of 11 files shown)


jcosentino11 (dismissed):

gdk/commands/test/config/InitConfiguration.py
```diff
@@ -10,6 +12,23 @@ def __init__(self, _args) -> None:
     def _get_otf_version(self):
```
MikeDombo, line 12: how are we getting OTF? Instead of only validating the semver, why not just check to see if the version requested actually exists? `100.0.0` is valid semver, but not a valid OTF version (at least not currently).

THIRD-PARTY-LICENSES
```diff
@@ -0,0 +1,31 @@
+
```
MikeDombo, line 1: where are the licenses for all other 3p?

gdk/commands/test/config/InitConfiguration.py
saranyailla: Updated code checks if the version tag exists.

Customers will have to provide the OTF version as they find it in the [releases](https://github.com/aws-greengrass/aws-greengrass-testing/releases). GDK will append `-SNAPSHOT` to that version for now as that's how all the testing jars have been so far.

MikeDombo (approved):

jcosentino11 (dismissed):

MikeDombo (approved):

urvashijain18 (approved):

PR #221, opened 2023-08-01
User recipe validation

nullPointer373:
**Issue #, if available:**

**Description of changes:**
Implemented JSON syntax validation for recipe files and added unit tests, including 
- checking recipe file existence
- opening the recipe file
- loading data from recipe file
- parsing syntax errors if any JSON decoder error occurs
- testing json validation functionalities in unit tests

This change is necessary to introduce recipe validation steps during the build stage to speed up the development cycle. It ensures that the recipe data is free from json syntax errors before proceeding with further processing.

Added unit tests to validate the functionality, and all unit tests (including the new ones) and integration tests passed successfully.

This modification lays the foundation for adding more validations, such as YAML syntax checks and semantic checks, within the validate_user_input_recipe.py file. The goal is to have the get_recipe() function in the file return a fully validated recipe data that can be used in the recipe transformer.

**Checklist:**


aws-kevinrickard:
I think I would prefer if you could integrate the changes into the recipe transformer / overall flow (such that you can show the new e2e output/behavior in an example run), but not a big deal. I'm not sure if there would be complications doing that in the current WIP state of the feature.

Looks good overall!

aws-kevinrickard (approved):

gdk/common/validate_user_input_recipe.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    user_recipe_file = get_user_input_recipe_file()
+
+    # validate the json recipe syntax
+    recipe_data = validate_recipe_syntax(user_recipe_file)
+
+    return recipe_data
+
+
+def get_user_input_recipe_file():
```
saranyailla, line 31: This piece here is redundant as we've an existing method to get the recipe file path in the GDKProject - https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/acaa624fae82bc8faf9e9ab726b710bdad0cd8bf/gdk/common/config/GDKProject.py#L11

gdk/common/validate_user_input_recipe.py
```diff
@@ -0,0 +1,84 @@
...
+
+    Parameters
+    ----------
+        None
+
+    Returns
+    -------
+        recipe_data: component's recipe as a dictionary object if the user-input recipe file is valid.
+    """
+    user_recipe_file = get_user_input_recipe_file()
```
saranyailla, line 23: GDK parses the recipes without adhering to the case (case-insensitive) https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/main/gdk/commands/component/transformer/BuildRecipeTransformer.py#L25

You may want to add the syntax and semantic validation on the already read file.

gdk/common/exceptions/error_messages.py
```diff
@@ -44,3 +51,34 @@
 
 # PUBLISH COMMAND
 PUBLISH_FAILED = "Failed to publish new version of component with the given configuration."
+
+# LIBRARY ERROR MESSAGES
+JSON_LIBRARY_ERROR_MESSAGES = {
```
saranyailla, line 56: Nit: You can create a separate file for this.

gdk/common/validate_user_input_recipe.py
saranyailla: Nit: IMO, this function should not be responsible for finding the recipe file path as that may not be the same every time. Eg. Build recipe file exists at the path `greengrass-build/recipes/recipe.yaml`. So, if we want to validate the build recipe file in future, this function won't be helpful.

We can rewrite the recipe validation in such a way that it can be extensible to the recipe path at any path. Something like

```
recipe_validator = RecipeValidator("/path/to/recipe")
recipe_validator.validate_syntax()
recipe_validator.validate_semantics() 
```

gdk/common/CaseInsensitive.py
```diff
@@ -84,7 +88,13 @@ def _read_from_yaml(self, file_path: Path) -> dict:
 
     def _read_from_json(self, file_path: Path) -> dict:
         with open(file_path, "r", encoding="utf-8") as f:
-            return json.loads(f.read())
+            try:
+                recipe_data = json.loads(f.read())
+                return recipe_data
+            except json.JSONDecodeError as err:
+                logging.error(f"Syntax error when parsing the recipe file: {file_path}")
+                parse_json_error(err)
```
saranyailla, line 96: Nit: `utils.parse_json_error(err)` is recommended for readability.

saranyailla (approved):

PR #222, opened 2023-08-07
feat: create json schema for user-input recipe

nullPointer373:
**Issue #, if available:**

**Description of changes:**
I have added a JSON schema file that defines the user-input recipe format and structure. The schema is based on the existing recipe definition in its Java package and its documentation, which allows for further semantic validation. 

The JSON schema is essential for validating user-input recipes and ensuring they conform to the expected format. By using the schema, we can enforce constraints on the recipe properties and catch any potential errors or invalid input early in the process.

I have tested the JSON schema by validating it against various recipes manually. And all unit tests (including newly added tests) and integration tests passed successfully.

One important aspect to consider is the naming convention for recipe properties. While the recipe follows the CamelCase naming convention for properties in its documentation, hands-on experimentation has revealed that the recipe properties can be case-insensitive when specified. To accommodate this behavior, I have deliberately made all property names in the JSON schema lowercase. During the validation process, the property names will be converted to lowercase in the data object before applying the schema. This approach ensures that the validation process is case-insensitive and can handle variations in property casing.

**Checklist:**

### DIFF
--- a/gdk/static/user_input_recipe_schema.json
+++ b/gdk/static/user_input_recipe_schema.json
@@ -1,0 +1,353 @@
+{
+    "$schema": "http://json-schema.org/draft/2020-12/schema",
+    "type": "object",
+    "properties": {
+        "recipeformatversion": {
+            "type": "string",
+            "enum": ["2020-01-25"]
+        },
+        "componentname": {
+            "type": "string",
+            "oneOf": [
+                {"pattern": "^[a-zA-Z0-9-_.]+$"},
+                {"enum": ["{COMPONENT_NAME}"]}
+            ],
+            "maxLength": 128
+        },
+        "componentversion": {
+            "type": "string",
+            "maxLength": 64,
+            "oneOf": [
+                {
+                    "pattern": "^(0|[1-9]\\d*)\\.(0|[1-9]\\d*)\\.(0|[1-9]\\d*)(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?"
+                },
+                {
+                    "enum": ["NEXT_PATCH", "{COMPONENT_VERSION}"]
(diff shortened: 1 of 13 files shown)


gdk/static/user_input_recipe_schema.json
```diff
@@ -0,0 +1,353 @@
+{
+    "$schema": "http://json-schema.org/draft/2020-12/schema",
+    "type": "object",
+    "properties": {
```
saranyailla, line 4: Are these properties case-insensitive?

gdk/static/user_input_recipe_schema.json
```diff
@@ -0,0 +1,353 @@
...
+                "additionalProperties": false
+            }
+        },
+        "lifecycle": {
+            "type": "object"
+        }
+    },
+    "required": ["recipeformatversion", "componentname"],
+    "additionalProperties": false
+}
```
saranyailla, line 353: can you also add tests for the same?

gdk/static/user_input_recipe_schema.json
```diff
@@ -0,0 +1,353 @@
...
+                },
+                "additionalProperties": false
+            }
+        },
+        "lifecycle": {
+            "type": "object"
+        }
+    },
+    "required": ["recipeformatversion", "componentname"],
+    "additionalProperties": false
```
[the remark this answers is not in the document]
nullPointer373: Regarding whether we should mark it as true, my suggestion is to set it to false. Enabling this option could lead to errors during the component creation process if there are additional keys beyond the predefined ones. To ensure compatibility with new recipe keys in the future, it's advisable to incorporate them into the schema. This way, the validation process can be extended to accommodate any potential changes.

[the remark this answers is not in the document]
saranyailla: Yes, that was my follow up question - if the component creation fails when undefined keys are added. In that case, I agree.

gdk/static/user_input_recipe_schema.json
nullPointer373: Based on my experience, the case doesn't matter for the recipe properties as long as the property name is accurate.

saranyailla (dismissed):

saranyailla (approved):

urvashijain18 (approved):

PR #223, opened 2023-08-09
feat: add yaml syntax validation

nullPointer373:
**Issue #, if available:**

**Description of changes:**
Implemented YAML syntax validation for recipe files and added unit tests, parsed syntax errors if any YAML error occurs.

It ensures that the recipe data is free from yaml syntax errors before proceeding with further processing.

Added unit tests to validate the functionality, and all unit tests (including the new ones) and integration tests passed successfully.

**Checklist:**

### DIFF
--- a/gdk/common/CaseInsensitive.py
+++ b/gdk/common/CaseInsensitive.py
@@ -6,7 +6,7 @@ from pathlib import Path
 import yaml
 from requests.structures import CaseInsensitiveDict as _CaseInsensitiveDict
 
-from gdk.common.utils import parse_json_error
+from gdk.common import utils
 
 
 class CaseInsensitiveDict(_CaseInsensitiveDict):
@@ -84,7 +84,13 @@ class CaseInsensitiveRecipeFile:
 
     def _read_from_yaml(self, file_path: Path) -> dict:
         with open(file_path, "r", encoding="utf-8") as f:
-            return yaml.safe_load(f.read())
+            try:
+                recipe_data = yaml.safe_load(f.read())
+                return recipe_data
+            except yaml.YAMLError as err:
+                logging.error(f"Syntax error when parsing the recipe file: {file_path}")
+                utils.parse_yaml_error(err)
+                sys.exit(1)
 
     def _read_from_json(self, file_path: Path) -> dict:
         with open(file_path, "r", encoding="utf-8") as f:
@@ -93,7 +99,7 @@ class CaseInsensitiveRecipeFile:
                 return recipe_data
             except json.JSONDecodeError as err:
                 logging.error(f"Syntax error when parsing the recipe file: {file_path}")
-                parse_json_error(err)
+                utils.parse_json_error(err)
                 sys.exit(1)
 
     def _write_to_json(self, file_path: Path, content: dict) -> None:
--- a/gdk/common/utils.py
+++ b/gdk/common/utils.py
@@ -165,12 +165,35 @@ def parse_json_error(err):
     msg = err.msg
 
     for err_msg, causes in syntax_error_message.JSON_LIBRARY_ERROR_MESSAGES.items():
-        if msg in err_msg:
+        if err_msg in msg:
             for cause in causes:
                 logging.info("\t " + cause)
             break
 
     logging.info("If none of the above is the cause, please review the overall JSON syntax and resolve any issues.")
+
+
+def parse_yaml_error(err):
+    """
+    Log and display a YAML syntax error message.
+
+    Parameters
+    ----------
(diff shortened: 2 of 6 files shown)


gdk/common/utils.py
```diff
@@ -165,14 +165,37 @@ def parse_json_error(err):
...
+    ----------
+    err : yaml.YAMLError
+        The YAML syntax error.
+
+    """
+    logging.error(f"{err}")
+    logging.info("This might be caused by one of the following reasons: ")
+    msg = err.problem
+
+    for err_msg, causes in syntax_error_message.YAML_LIBRARY_ERROR_MESSAGES.items():
```
saranyailla, line 190: Have you checked if there are (inner) methods of the `yaml` library that we can use to get the `msg` that looks similar to the ones in the `YAML_LIBRARY_ERROR_MESSAGES`?

gdk/common/utils.py
```diff
@@ -165,14 +165,37 @@ def parse_json_error(err):
...
+        if err_msg in msg:
```
saranyailla, line 191: What if the newer version of `yaml` library gives us a new error message that isn't part of the list we're maintaining? In that case, we don't want the  following info log.

gdk/common/utils.py
```diff
@@ -165,14 +165,37 @@ def parse_json_error(err):
...
+            for cause in causes:
+                logging.info("\t " + cause)
+            break
+
+    logging.info("If none of the above is the cause, please review the overall YAML syntax and resolve any issues.")
```
saranyailla, line 196: Can move this to L 194?

gdk/common/CaseInsensitive.py
```diff
@@ -84,7 +84,13 @@ def _read(self, file_path):
...
     def _read_from_yaml(self, file_path: Path) -> dict:
         with open(file_path, "r", encoding="utf-8") as f:
-            return yaml.safe_load(f.read())
+            try:
+                recipe_data = yaml.safe_load(f.read())
+                return recipe_data
+            except yaml.YAMLError as err:
+                logging.error(f"Syntax error when parsing the recipe file: {file_path}")
+                utils.parse_yaml_error(err)
+                sys.exit(1)
```
saranyailla, line 93: raise Exception here with the error log message you have.

gdk/common/utils.py
nullPointer373: I've looked into it, and it seems that the YAML library doesn't provide inner methods with error messages that can be accessed publicly. Additionally, there doesn't appear to be a clear documentation that lists all possible error messages.

gdk/common/utils.py
nullPointer373: It's quite uncommon for the yaml library to alter its error messages. However, if a new error message does arise that isn't on our maintained list, the info log won't include potential causes for that specific message. Rest assured, the error message itself will still be displayed, allowing developers to identify and troubleshoot the issue effectively.

gdk/common/CaseInsensitive.py
nullPointer373: Using raise Exception would trigger a lengthy traceback, which might not be the most user-friendly. Given that we can already pinpoint that the exception is due to a recipe syntax error and provide clear information about the problematic file, do we still need to raise the exception in this context?

gdk/common/utils.py
saranyailla: I was referring to the info log in L196.

nullPointer373: I can reposition the statement "If none of the above is the cause" within the if clause so that it's displayed only when potential causes are logged.

gdk/common/CaseInsensitive.py
saranyailla: Yes, we're moving towards printing the full stack trace with the recent versions of GDK. Here, you can raise exception without creating a new one and keep it consistent with how rest of the GDK works.

gdk/common/utils.py
```diff
@@ -161,16 +161,17 @@ def parse_json_error(err):
...
 
-    logging.info("This might be caused by one of the following reasons: ")
     msg = err.msg
 
     for err_msg, causes in syntax_error_message.JSON_LIBRARY_ERROR_MESSAGES.items():
         if err_msg in msg:
+            logging.info("This might be caused by one of the following reasons: ")
             for cause in causes:
                 logging.info("\t " + cause)
+            logging.info("If none of the above is the cause,")
```
saranyailla, line 171: Nit: IMO, this line can be removed completely as the log below works well in both the cases.

saranyailla (approved):

PR #226, opened 2023-08-10
feat: wizard input validator and test cases 

Jacksonzhang34:
**Issue #, if available:**

**Description of changes:**
wizard input validator and test cases 
In order to valid the input responses of the customers to the prompt questions about field key values 
unit testing 
Todo: Next PR will include validators along with test cases for build options, the publish configuration, and gdk version
**Checklist:**

### DIFF
--- a/gdk/wizard/WizardChecker.py
+++ b/gdk/wizard/WizardChecker.py
@@ -1,4 +1,4 @@
-from gdk.wizard.commons.fields import Fields
+from gdk.wizard.ConfigEnum import ConfigEnum
 import re
 import ast
 
@@ -6,7 +6,7 @@ import ast
 class WizardChecker:
     def is_valid_input(self, input_value, field):
         """
-        Prompts the user of all the optional fields of the gdk config file and updates the
+        Prompts the user of all the optional ConfigEnum of the gdk config file and updates the
         field_map if their answer is valid  as the user answers the question to each prompt
 
         Parameters
@@ -20,15 +20,10 @@ class WizardChecker:
         """
 
         switch = {
-            Fields.AUTHOR: self.check_author,
-            Fields.VERSION: self.check_version,
-            Fields.CUSTOM_BUILD_COMMAND: self.check_custom_build_command,
-            Fields.BUILD_SYSTEM: self.check_build_system,
-            Fields.BUILD_OPTIONS: self.check_build_options,
-            Fields.BUCKET: self.check_bucket,
-            Fields.REGION: self.check_region,
-            Fields.PUBLISH_OPTIONS: self.check_publish_options,
-            Fields.GDK_VERSION: self.check_gdk_version,
+            ConfigEnum.AUTHOR: self.check_author,
+            ConfigEnum.VERSION: self.check_version,
+            ConfigEnum.CUSTOM_BUILD_COMMAND: self.check_custom_build_command,
+            ConfigEnum.BUILD_SYSTEM: self.check_build_system,
         }
         return switch.get(field)(input_value)
 
--- a/gdk/wizard/WizardChecker.py
+++ b/gdk/wizard/WizardChecker.py
@@ -1,0 +1,83 @@
+from gdk.wizard.commons.fields import Fields
+import re
+import ast
+
+
+class WizardChecker:
+    def is_valid_input(self, input_value, field):
+        """
+        Prompts the user of all the optional fields of the gdk config file and updates the
+        field_map if their answer is valid  as the user answers the question to each prompt
+
+        Parameters
+        ----------
+            input_value(string): The value entered by the user for the field
+            field(string): The field name of the field in the gdk config file
+
+        Returns
+        -------
+            boolean: True if the input value is valid for the field, False otherwise.
+        """
+
+        switch = {
+            Fields.AUTHOR: self.check_author,
+            Fields.VERSION: self.check_version,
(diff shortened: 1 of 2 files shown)


gdk/wizard/WizardChecker.py
```diff
@@ -0,0 +1,78 @@
...
+            ConfigEnum.AUTHOR: self.check_author,
+            ConfigEnum.VERSION: self.check_version,
+            ConfigEnum.CUSTOM_BUILD_COMMAND: self.check_custom_build_command,
+            ConfigEnum.BUILD_SYSTEM: self.check_build_system,
+        }
+        return switch.get(field)(input_value)
+
+    def check_author(self, input_value):
+        # input must be a non-empty string
+        return isinstance(input_value, str) and len(input_value) > 0
```
saranyailla, line 32: What about max length? This limit should be same as what Greengrass supports.

gdk/wizard/WizardChecker.py
```diff
@@ -0,0 +1,78 @@
...
+        -------
+            boolean: True if the input value is valid for the field, False otherwise.
+        """
+
+        switch = {
+            ConfigEnum.AUTHOR: self.check_author,
+            ConfigEnum.VERSION: self.check_version,
+            ConfigEnum.CUSTOM_BUILD_COMMAND: self.check_custom_build_command,
+            ConfigEnum.BUILD_SYSTEM: self.check_build_system,
+        }
```
saranyailla, line 27: Nit: Move this `switch` out into `__init__` of `WizardChecker` as we don't need to construct `switch` every time a caller uses `is_valid_input()`.

saranyailla: Nit: Also, all of these checks are returning boolean values. So, you can rename `check_author` to `is_valid_author`.

gdk/wizard/WizardChecker.py
```diff
@@ -0,0 +1,78 @@
...
+    def check_custom_build_command(self, input_value):
+        # input must be a non-empty string or a non-empty list of strings
+
+        # since input_value will always be a string (b/c its a CLI input), need to try and convert it
+        # to a list of strings. If it fails, then the input is not valid. Then command_list a non-empty
+        # list of strings
+
+        # required if build_system is custom, exit the wizard with error message
+        # second approach revert to build_system to zip
+        # or infinite loop
```
saranyailla, line 63: Nit: Use `"""` for multi-line comments.

gdk/wizard/WizardChecker.py
Jacksonzhang34: Yes i just moved it. Should i change all the checks to is_valid ?

gdk/wizard/WizardChecker.py
Jacksonzhang34: I just based my checks of what the schema specified. I looked at here https://docs.aws.amazon.com/greengrass/v2/developerguide/gdk-cli-configuration-file.html#gdk-config-format
to see the speifications for each field was but I didnt see anything about the max length

gdk/wizard/WizardChecker.py
```diff
@@ -0,0 +1,78 @@
...
+
+        try:
+            input_list = ast.literal_eval(input_value)
+            if isinstance(input_list, list):
+                return (
+                    all(isinstance(item, str) and item.strip() for item in input_list)
+                    and len(input_list) > 0
+                )
+            else:
+                return False
```
saranyailla, line 76: Why are we returning `False` if it doesn't evaluate to a `list`?

gdk/wizard/WizardChecker.py
```diff
@@ -0,0 +1,78 @@
...
+        except (ValueError, SyntaxError):
+            return len(input_value) > 0
```
saranyailla, line 78: As this can raise all these kinds of errors here https://docs.python.org/3/library/ast.html#ast.literal_eval, use `Exception` to catch all and also return `False` without verifying the input size here.

gdk/wizard/WizardChecker.py
Jacksonzhang34: The reasoning is input_list = ast.literal_eval(input_value) will try to parse the input string. And if that string is not a list or set ({}) then there will be a parsing error so there will be an exception, in which case it is assumed to be just any string that the user has inputted, so I just check if the length is greater than 0

gdk/wizard/WizardChecker.py
Jacksonzhang34: alright Ill replace except (ValueError, SyntaxError): with except Exception:

gdk/wizard/WizardChecker.py
saranyailla: From what I see [here](https://docs.python.org/3/library/ast.html#:~:text=the%20string%20or%20node%20provided%20may%20only%20consist%20of%20the%20following%20python%20literal%20structures%3A%20strings%2C%20bytes%2C%20numbers%2C%20tuples%2C%20lists%2C%20dicts%2C%20sets%2C%20booleans%2C%20none%20and%20ellipsis), it can be evaluated to any one of these strings, bytes, numbers, tuples, lists, dicts, sets, booleans, None and Ellipsis.

saranyailla (approved):

saranyailla (approved):

urvashijain18 (approved):

PR #228, opened 2023-08-11
feat: Wizard prompter 

Jacksonzhang34:
**Issue #, if available:**

**Description of changes:**
Wizard prompter
prompters users to populate config key values 
This change has not yet been tested

**Checklist:**

### DIFF
--- a/gdk/wizard/Prompter.py
+++ b/gdk/wizard/Prompter.py
@@ -32,7 +32,7 @@ class Wizard:
         self.checker = WizardChecker()
         self.parser = argparse.ArgumentParser()
 
-    def prompter(self, field, value, required, max_attempts=3):
+    def prompter(self, field, required, max_attempts=3):
         """
         Prompts the user for a value of a given field key
 
@@ -47,6 +47,7 @@ class Wizard:
         -------
             string: the value for the field key "field"
         """
+        current_field_value = self.data.get_field(field)
         require = "REQUIRED " if required else "OPTIONAL "
         link = "https://docs.aws.amazon.com/greengrass/v2/developerguide/gdk-cli-configuration-file.html#gdk-config-format"
         for attempt in range(1, max_attempts + 1):
@@ -59,7 +60,9 @@ class Wizard:
             args = self.parser.parse_args(
                 [
                     f"--{parser_argument}",
-                    self.interactive_prompt(field.value.key, value, require),
+                    self.interactive_prompt(
+                        parser_argument, current_field_value, require
+                    ),
                 ]
             )
             response = getattr(args, parser_argument).strip()
@@ -77,7 +80,7 @@ class Wizard:
             # the same as the current value, return the current value, then return the same value
             if (
                 response == field.value.default
-                or response == value
+                or response == current_field_value
                 or self.checker.is_valid_input(response, field)
             ):
                 return response
@@ -91,8 +94,8 @@ class Wizard:
                 "You have failed to enter a valid custom build command. Exiting wizard..."
             )
 
-        print("Exceeded maximum attempts. Assuming default response.")
-        return value
+        print("Exceeded maximum attempts. Assuming current or default response.")
+        return current_field_value
 
     def change_configuration(self, field_key, max_attempts=3):
         """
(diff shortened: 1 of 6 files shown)


gdk/wizard/Prompter.py
```diff
@@ -0,0 +1,259 @@
...
+                "message": f"Current value of the {require}{field} is:",
+                "default": f"{value}",
+            }
+        ]
+        try:
+            answer = prompt(questions)
+            return answer["user_input"]
+        except (KeyError, TypeError):
+            self.utils.write_to_config_file(self.field_dict, self.project_config_file)
+            sys.exit("Wizard interrupted. Exiting...")
```
saranyailla, line 162: raise Exception here.

gdk/wizard/Prompter.py
```diff
@@ -0,0 +1,259 @@
...
+                        parser_argument, current_field_value, require
+                    ),
+                ]
+            )
+            response = getattr(args, parser_argument).strip()
+
+            # only way customer is asked about custom_build_command if they have custom build system
+            # in which case they must specify a custom build command that is not None
+            if field == ConfigEnum.CUSTOM_BUILD_COMMAND:
+                if response == "None":
```
saranyailla, line 73: What happens when the input is an empty string on an empty [] object?

gdk/wizard/Prompter.py
Jacksonzhang34: so instead of write_to_config_file and exiting wizard raise an Exception instead?

gdk/wizard/Prompter.py
```diff
@@ -0,0 +1,259 @@
...
+                    print(
+                        f"Attempt {attempt}/{max_attempts}: Must Specify a custum build command.\nPlease vist: {link}"
+                    )
+                    continue
+
+            # if customer input response is the default value or the customer input response is
+            # the same as the current value, return the current value, then return the same value
+            if (
+                response == field.value.default
+                or response == current_field_value
```
saranyailla, line 83: I don't think we need these two checks. Just check if the response `is_valid_input`.

gdk/wizard/Prompter.py
```diff
@@ -0,0 +1,259 @@
...
+                f"Attempt {attempt}/{max_attempts}: Invalid response. Please try again.\nPlease vist: {link}"
+            )
+
+        if field == ConfigEnum.CUSTOM_BUILD_COMMAND:
+            self.utils.write_to_config_file(self.field_dict, self.project_config_file)
+            sys.exit(
+                "You have failed to enter a valid custom build command. Exiting wizard..."
+            )
+
+        print("Exceeded maximum attempts. Assuming current or default response.")
```
saranyailla, line 97: Use `logging` instead of `print`.

gdk/wizard/Prompter.py
```diff
@@ -0,0 +1,259 @@
...
+                return response
+            print(
+                f"Attempt {attempt}/{max_attempts}: Invalid response. Please try again.\nPlease vist: {link}"
+            )
+
+        if field == ConfigEnum.CUSTOM_BUILD_COMMAND:
+            self.utils.write_to_config_file(self.field_dict, self.project_config_file)
+            sys.exit(
+                "You have failed to enter a valid custom build command. Exiting wizard..."
+            )
```
saranyailla, line 95: Why do we need this?

gdk/wizard/Prompter.py
Jacksonzhang34: I see, I forgot to account for that. My thought is to also check len(response) < 1 what do you think of that

gdk/wizard/Prompter.py
Jacksonzhang34: Got it

gdk/wizard/Prompter.py
Jacksonzhang34: got it

gdk/wizard/Prompter.py
Jacksonzhang34: gotcha so if the program is still executing outside of the for loop it means that the response from the user is still not valid after the max number of tries. Thus, if its the case that wizard is prompting for custom build command then, we must exit the wizard instead of reprompting, if we're not prompting the wizard then we just take the current valid already in the field_map

gdk/wizard/Prompter.py
```diff
@@ -0,0 +1,259 @@
...
+
+        self.utils = WizardConfigUtils()
+        self.project_config_file = self.utils.get_project_config_file()
+        self.field_dict = self.utils.read_from_config_file()
+
+        self.data = WizardData(self.field_dict)
+        self.checker = WizardChecker()
+        self.parser = argparse.ArgumentParser()
+
+    def prompter(self, field, required, max_attempts=3):
```
saranyailla, line 35: This is for later - should we also pass help text as an arg to this method. 

For eg. when user fails to provide a valid input for `version`, the help log should be something like 
> Please input a valid version for the component in semver format. Eg. 1.0.0

Jacksonzhang34: I see yes definitely. I'll put it in the TODOs list

gdk/wizard/Prompter.py
```diff
@@ -0,0 +1,259 @@
...
+                response == field.value.default
+                or response == current_field_value
+                or self.checker.is_valid_input(response, field)
+            ):
+                return response
+            print(
+                f"Attempt {attempt}/{max_attempts}: Invalid response. Please try again.\nPlease vist: {link}"
+            )
+
+        if field == ConfigEnum.CUSTOM_BUILD_COMMAND:
```
saranyailla, line 91: Add this validation to the checker file `is_valid_custom_build_command`.

gdk/wizard/Prompter.py
saranyailla: Add this validation to the checker file is_valid_custom_build_command.

saranyailla: You can then remove L72-L77.

saranyailla (approved):

urvashijain18 (approved):

PR #235, opened 2023-08-30
feat: add recipe file size validation

nullPointer373:
**Issue #, if available:**

**Description of changes:**
Incorporated a recipe file size validation to ensure the file size remains within the permissible limit of 16KB. Added a verification process for the validity and compatibility of the RecipeFormatVersion, with enhanced error messages for better customer guidance.

This change is crucial to establish a comprehensive recipe validation mechanism, encompassing both size and content checks. By doing so, we enhance the reliability of recipes for downstream processes.

Introduced new unit tests that successfully passed, affirming the functionality. Thoroughly validated the implementation locally to ensure it behaves as intended.

The file size validation is conducted prior to processing the recipe file's contents. Below are illustrative outputs for scenarios:

- Recipe with an invalid file size.
[2023-08-30 11:50:58] INFO - Building the component 'com.example.HelloWorld' with the given project configuration.
[2023-08-30 11:50:58] INFO - Using 'zip' build system to build the component.
[2023-08-30 11:50:58] WARNING - This component is identified as using 'zip' build system. If this is incorrect, please exit and specify custom build command in the 'gdk-config.json'.
[2023-08-30 11:50:58] INFO - Validating the recipe file /Users/yifanzyf/Desktop/HelloWorld/recipe.json
[2023-08-30 11:50:58] ERROR - The size of the recipe exceeds the maximum allowed size. Please ensure the recipe size does not exceed 16 KB.
[2023-08-30 11:50:58] ERROR - Failed to build the component with the given project configuration.

- Recipe missing the RecipeFormatVersion.
[2023-08-30 12:01:27] INFO - Building the component 'com.example.HelloWorld' with the given project configuration.
[2023-08-30 12:01:27] INFO - Using 'zip' build system to build the component.
[2023-08-30 12:01:27] WARNING - This component is identified as using 'zip' build system. If this is incorrect, please exit and specify custom build command in the 'gdk-config.json'.
[2023-08-30 12:01:27] INFO - Validating the recipe file /Users/yifanzyf/Desktop/HelloWorld/recipe.json
[2023-08-30 12:01:27] ERROR - Recipe validation failed for 'RecipeFormatVersion'. This field is required but missing from the recipe. Please correct it and try again.
[2023-08-30 12:01:27] ERROR - Failed to build the component with the given project configuration.

- Recipe containing an unsupported RecipeFormatVersion.
[2023-08-30 12:03:50] INFO - Building the component 'com.example.HelloWorld' with the given project configuration.
[2023-08-30 12:03:50] INFO - Using 'zip' build system to build the component.
[2023-08-30 12:03:50] WARNING - This component is identified as using 'zip' build system. If this is incorrect, please exit and specify custom build command in the 'gdk-config.json'.
[2023-08-30 12:03:50] INFO - Validating the recipe file /Users/yifanzyf/Desktop/HelloWorld/recipe.json
[2023-08-30 12:03:50] ERROR - Recipe validation failed for: 'RecipeFormatVersion: 2023-01-25'.
[2023-08-30 12:03:50] ERROR - The provided RecipeFormatVersion '2023-01-25' is not supported in this gdk version. Please ensure that it is a valid RecipeFormatVersion compatible with the gdk, and refer to the list of supported RecipeFormatVersion: ['2020-01-25'].
[2023-08-30 12:03:50] ERROR - Failed to build the component with the given project configuration.

**Checklist:**

### DIFF
--- a/gdk/commands/component/transformer/BuildRecipeTransformer.py
+++ b/gdk/commands/component/transformer/BuildRecipeTransformer.py
@@ -27,14 +27,23 @@ class BuildRecipeTransformer:
         return self._s3_client
 
     def transform(self, build_folders):
-        component_recipe = CaseInsensitiveRecipeFile().read(self.project_config.recipe_file)
         logging.info(f"Validating the recipe file {self.project_config.recipe_file}")
+
+        # Validate the size of the recipe file before processing its content.
+        if not utils.valid_recipe_file_size(self.project_config.recipe_file):
+            logging.error("The size of the recipe exceeds the maximum allowed size. Please ensure the recipe size "
+                          "does not exceed 16 KB.")
+            raise Exception(error_messages.RECIPE_SIZE_INVALID.format(self.project_config.recipe_file))
+
+        component_recipe = CaseInsensitiveRecipeFile().read(self.project_config.recipe_file)
         try:
             validator = RecipeValidator(component_recipe)
+            validator.validate_recipe_format_version()
             validator.validate_semantics()
             validator.validate_input()
         except jsonschema.exceptions.ValidationError as err:
             raise Exception(error_messages.RECIPE_FILE_INVALID.format(self.project_config.recipe_file, err.message))
+
         self.update_component_recipe_file(component_recipe, build_folders)
         self.create_build_recipe_file(component_recipe)
 
--- a/gdk/commands/component/transformer/BuildRecipeTransformer.py
+++ b/gdk/commands/component/transformer/BuildRecipeTransformer.py
@@ -31,8 +31,7 @@ class BuildRecipeTransformer:
 
         # Validate the size of the recipe file before processing its content.
         if not utils.valid_recipe_file_size(self.project_config.recipe_file):
-            logging.error("The size of the recipe exceeds the maximum allowed size. Please ensure the recipe size "
-                          "does not exceed 16 KB.")
+            logging.error(error_messages.RECIPE_SIZE_INVALID.format(self.project_config.recipe_file))
             raise Exception(error_messages.RECIPE_SIZE_INVALID.format(self.project_config.recipe_file))
 
         component_recipe = CaseInsensitiveRecipeFile().read(self.project_config.recipe_file)
--- a/gdk/common/RecipeValidator.py
+++ b/gdk/common/RecipeValidator.py
@@ -28,6 +28,9 @@ class RecipeValidator:
 
     Methods
     -------
+    validate_recipe_format_version()
+        Validates that the provided RecipeFormatVersion in the recipe is valid and compatible with the gdk.
+
     validate_semantics()
         Validates the semantics of the component recipe against the Greengrass component recipe schema.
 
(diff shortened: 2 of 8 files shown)


gdk/common/RecipeValidator.py
```diff
@@ -49,6 +52,33 @@ def __init__(self, recipe_source):
...
+        recipe_format_version = self.recipe_data.get("recipeformatversion")
+        if not recipe_format_version:
+            logging.error("Recipe validation failed for 'RecipeFormatVersion'. This field is required but missing "
+                          "from the recipe. Please correct it and try again.")
+            raise Exception("The recipe file is invalid. The 'RecipeFormatVersion' field is mandatory in the recipe.")
+
+        supported_recipe_version = ["2020-01-25"]
+        if recipe_format_version not in supported_recipe_version:
+            logging.error(f"Recipe validation failed for: 'RecipeFormatVersion: {recipe_format_version}'.")
+            logging.error(f"The provided RecipeFormatVersion '{recipe_format_version}' is not supported in this gdk "
```
aws-kevinrickard, line 76: I'm thinking this might be better off as a warning?

Because depending on the changes to the recipe format version, it might not break an older version of gdk being able to work with it. This leaves the option for the customer to continue using their current gdk without upgrading if they want to take that risk. 

However I wouldn't want a customer to miss the warning and be surprised when something isn't working as expected later on. So if we do go the warning route maybe we can make the message extra noticeable somehow; potentially with some ascii art (if that fits in with GDK/GG design).

I'm open to hearing more opinions on this, I can see tradeoffs either way.

gdk/common/utils.py
```diff
@@ -226,6 +227,11 @@ def parse_json_schema_errors(error):
             logging.info(f"\t {fix}")
 
 
+def valid_recipe_file_size(file_path):
+    file_size = os.path.getsize(file_path)
+    return file_size < 16234    # max allowed byte
```
aws-kevinrickard, line 232: nit: Rather than noting the purpose of the literal in a comment, I think it's better to write it as a constant with an appropriate name. Because a comment is more likely to be ignored and not updated if the code is changed later, and this way the code reads more naturally in my opinion.

 e.g. 

```
MAX_RECIPE_FILE_SIZE_BYTES = 16234
file_size_bytes = os.path.getsize(file_path)
return file_size_bytes < MAX_RECIPE_FILE_SIZE_BYTES
```

gdk/common/utils.py
```diff
@@ -226,6 +227,11 @@ def parse_json_schema_errors(error):
             logging.info(f"\t {fix}")
 
 
+def valid_recipe_file_size(file_path):
```
aws-kevinrickard, line 230: Why 16234?

I think it's worth double checking we have the correct value here because kilobyte is a slightly ambiguous term, it can be defined as either 1000 bytes (kB) or 1024 bytes (KiB / KB). I want to make sure we're using the correct one. 

And more importantly using either of these definitions for 16 kilobyte does not result in 16234, 
i.e. 16 kB => 16000, 16 KB => 16384.

gdk/commands/component/transformer/BuildRecipeTransformer.py
```diff
@@ -27,14 +27,23 @@ def _get_s3_client(self, _region):
         return self._s3_client
 
     def transform(self, build_folders):
-        component_recipe = CaseInsensitiveRecipeFile().read(self.project_config.recipe_file)
         logging.info(f"Validating the recipe file {self.project_config.recipe_file}")
+
+        # Validate the size of the recipe file before processing its content.
+        if not utils.valid_recipe_file_size(self.project_config.recipe_file):
+            logging.error("The size of the recipe exceeds the maximum allowed size. Please ensure the recipe size "
```
aws-kevinrickard, line 34: Instead of defining a similar error message both here and in error_messages, can we re-use the same error message for both the log and the exception? I think that would be easier to maintain.

Similar concern for the other cases of log and then exception down below, although those cases don't use error_messages.

gdk/common/utils.py
nullPointer373: I tested a large file with the GG API for component creation and found that the maximum allowed size is 16233 bytes. Any size exceeding 16233 bytes, such as 16234 bytes, will result in a failure, accompanied by the error message: "Invalid Input: Size of the recipe can't exceed 16KB". Although 16233 might seem like a magic number, I chose to use it due to the reasons mentioned above. 

Perhaps we could consider employing the threshold of 16kB (16000 bytes) in this context.

aws-kevinrickard: As discussed offline, 16kB or 16000 bytes should be a safe choice with clear boundaries for the customer

aws-kevinrickard (approved):

saranyailla (approved):

PR #236, opened 2023-09-06
feat: update recipe schema to recommended CamelCase

nullPointer373:
**Issue #, if available:**

**Description of changes:**
Updated the recipe schema to CamelCase as documented in the [AWS IoT Greengrass component recipe reference](https://docs.aws.amazon.com/greengrass/v2/developerguide/component-recipe-reference.html) and changed the corresponding validation to align with the recipe schema.

This change is necessary to improve the error logging message and make it more informative, in line with user input expectations.

New unit tests were added, and all tests passed successfully. Additionally, the manual testing was conducted to ensure it functions as expected.

**Checklist:**

### DIFF
--- a/gdk/common/RecipeValidator.py
+++ b/gdk/common/RecipeValidator.py
@@ -50,7 +50,7 @@ class RecipeValidator:
 
         """
         self.recipe_source = recipe_source
-        self.recipe_data = self._convert_keys_to_lowercase(self._load_recipe())
+        self.recipe_data = self._convert_keys_to_camelcase(self._load_recipe())
 
     def validate_recipe_format_version(self):
         """
@@ -64,7 +64,7 @@ class RecipeValidator:
             If the RecipeFormatVersion field is missing.
 
         """
-        recipe_format_version = self.recipe_data.get("recipeformatversion")
+        recipe_format_version = self.recipe_data.get("RecipeFormatVersion")
         if not recipe_format_version:
             err_msg = "Recipe validation failed for 'RecipeFormatVersion'. This field is required but missing " \
                       "from the recipe. Please correct it and try again."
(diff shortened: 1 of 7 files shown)


gdk/common/RecipeValidator.py
```diff
@@ -260,3 +260,75 @@ def _convert_keys_to_lowercase(self, input_dict):
...
+        elif isinstance(input_data, list):
+            result_list = []
+            for item in input_data:
+                result_list.append(self._convert_keys_to_camelcase(item))
+            return result_list
+        else:
+            return input_data
+
+    RECIPE_PROPERTY_CASE_MAPPING = {
+        "recipeformatversion": "RecipeFormatVersion",
```
aws-kevinrickard, line 294: Can we generate this from the recipe schema, rather than re-defining these here? I think it should be possible, by reading all properties from the schema (which would get the camelcase data), and then converting that to lowercase and assembling it into a map like this. This would make maintenance on the schema easier, by not needing to also update this.

gdk/common/RecipeValidator.py
```diff
@@ -249,7 +249,7 @@ def _convert_keys_to_lowercase(self, input_dict):
             The input dictionary or list.
 
         Returns
-        -------
+        -------0
```
aws-kevinrickard, line 252: small typo

aws-kevinrickard (changes requested):

aws-kevinrickard (approved):

saranyailla (approved):

gdk/common/RecipeValidator.py
```diff
@@ -290,45 +290,57 @@ def _convert_keys_to_camelcase(self, input_data):
...
+            Parameters
+            ----------
+            data : dict or list
+                A portion of the recipe schema.
+
+            """
+            if isinstance(data, dict):
+                for key, value in data.items():
+                    # eliminate the keywords of Json Schema
+                    if key not in syntax_error_message.JSON_SCHEMA_VALIDATION_KEYWORDS.keys() and key != '$schema':
```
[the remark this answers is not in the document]
nullPointer373: My concern regarding this approach is that not all recipe properties necessarily fall directly under the 'properties' schema keyword. Some properties may be structured differently, residing under data structures specified by schema keywords like 'oneOf' or 'anyOf.' In such cases, we would still need to rely on specific keywords and address other data structures. 

Also, I think the current approach of extracting all keys and then eliminating schema keywords is a more unified and maintainable solution. If schema keywords change, we can adjust the list of keywords. In contrast, the alternative approach might require changing the implementation for different keywords, and it might add complexity to the property retrieval.

PR #243, opened 2023-09-19
feat: add recipe size validation during build stage

patrzhan:
**Issue #, if available:**

**Description of changes:**
Add recipe size validation to the build stage of the GDK. When the command `gdk component build` is run, it will check that the recipe file size is less than 16 kB. If it is over the limit, the command will fail with an exception and exit without completing the build.

Greengrass only supports recipe files up to 16 kB so adding validation will allow GDK users to identify this potential point of failure while they try to build their project rather than later during a component deployment.

Added unit tests for the added function, added integration test with a case where the provided file size is larger than 16kB and build is attempted to be ran.

**Checklist:**


hancin (approved):

gdk/commands/component/transformer/BuildRecipeTransformer.py
```diff
@@ -22,6 +23,12 @@ def _get_s3_client(self, _region):
         return self._s3_client
 
     def transform(self, build_folders):
+        logging.info(f"Validating the file size of recipe {self.project_config.recipe_file}")
+        # Validate the size of the recipe file before processing its content.
+        if not utils.valid_recipe_file_size(self.project_config.recipe_file):
```
saranyailla, line 28: Nit: Rename it to reflect a boolean method - `is_recipe_size_valid`?

We'll have to do the same check right before publishing the component as well as that's the final recipe that we're going to use to create the component.

tests/gdk/common/test_utils.py
```diff
@@ -181,3 +181,15 @@ def test_cli_version_check_latest_available(mocker):
 )
 def test_get_next_patch_version(version):
     assert utils.get_next_patch_version(version) == "1.0.1"
+
+
+def test_valid_recipe_size(mocker):
+    mocker.patch("os.path.getsize", return_value=1000)
+    result = utils.valid_recipe_file_size('small_recipe.json')
+    assert result
```
saranyailla, line 189: Nit: create/mock a file with contents size less than 16KB and then test if it is valid. Same for the test below.

gdk/common/exceptions/error_messages.py
```diff
@@ -8,6 +8,7 @@
 )
 PROJECT_CONFIG_FILE_INVALID = "Project configuration file '{}' is invalid. Please correct its format and try again. Error: {} "
 CLI_MODEL_FILE_NOT_EXISTS = "Model validation failed. CLI model file doesn't exist."
+RECIPE_SIZE_INVALID = "The input recipe file '{}' has an invalid size. Please make sure it does not exceed 16kB and try again."
```
saranyailla, line 11: Nit: Should we want to log its current size instead of calling it 'invalid'?

gdk/commands/component/transformer/PublishRecipeTransformer.py
```diff
@@ -13,6 +14,14 @@ def __init__(self, _project_config: ComponentPublishConfiguration) -> None:
...
     def transform(self):
         recipe_path = Path(self.project_config.gg_build_recipes_dir).joinpath(self.project_config.recipe_file.name)
+
+        logging.info(f"Validating the file size of the built recipe {recipe_path}")
+        # Validate the size of the recipe file before processing its content.
+        valid_file_size, input_recipe_file_size = utils.is_recipe_size_valid(recipe_path)
+        if not valid_file_size:
+            logging.error(BUILT_RECIPE_SIZE_INVALID.format(self.project_config.recipe_file, input_recipe_file_size))
+            raise Exception(BUILT_RECIPE_SIZE_INVALID.format(self.project_config.recipe_file, input_recipe_file_size))
+
```
saranyailla, line 24: https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/242cf804f24b6f4292c84c07bcaa99ca2154fdb6/gdk/commands/component/transformer/PublishRecipeTransformer.py#L94 we want to add this here, after updating and before creating the recipe file.

saranyailla (approved):

hancin (approved):

PR #244, opened 2023-09-21
feat: better logging for recipe syntax failures during build

patrzhan:
**Issue #, if available:**

**Description of changes:**
Add better logging during build stage for recipe syntax validation. Before, when a build fails due to a recipe syntax issue it just prints the error that there was some problem during the build. Now, it will notify the user that there was an issue with the recipe syntax and link to the component recipe docs for reference.

Provides more clarity for what is going wrong.

Added unit tests for recipes with a common syntax error.

**Checklist:**

### DIFF
--- a/tests/gdk/common/test_CaseInsensitive.py
+++ b/tests/gdk/common/test_CaseInsensitive.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import tempfile
 from pathlib import Path
 from unittest import TestCase
@@ -14,6 +15,10 @@ class CaseInsensitiveRecipeFileTest(TestCase):
     @pytest.fixture(autouse=True)
     def __inject_fixtures(self, mocker):
         self.mocker = mocker
+
+    @pytest.fixture(autouse=True)
+    def caplog(self, caplog):
+        self.caplog = caplog
 
     def test_read_json(self):
         json_file = Path(".").joinpath("tests/gdk/static/project_utils").joinpath("valid_component_recipe.json").resolve()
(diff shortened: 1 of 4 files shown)


saranyailla (dismissed):

tests/gdk/common/test_CaseInsensitive.py
```diff
@@ -37,6 +42,24 @@ def test_read_yaml(self):
...
+        json_file = Path(".").joinpath("tests/gdk/static/project_utils").joinpath("recipe_missing_comma.json").resolve()
+        with pytest.raises(Exception) as e:
+            CaseInsensitiveRecipeFile().read(json_file)
+        logs = self.caplog.text
+        assert "For information and examples regarding component recipes refer to the docs here" in logs
+        assert "Expecting ',' delimiter: line 6 column 3" in str(e)
+
+    def test_read_yaml_with_syntax_err(self):
+        self.caplog.set_level(logging.ERROR)
+        json_file = Path(".").joinpath("tests/gdk/static/project_utils").joinpath("invalid_component_recipe.yaml").resolve()
```
hancin, line 56: Since we didn't create this file for this test, are we sure it will remain consistent for this test?

hancin (dismissed):

tests/gdk/common/test_CaseInsensitive.py
hancin: nit: it's also not a `json_file`

patrzhan: Good catch, I'll update the variable name.

tests/gdk/common/test_CaseInsensitive.py
```diff
@@ -37,6 +42,24 @@ def test_read_yaml(self):
         assert "uri" in case_insensitive_recipe["Manifests"][0]["Artifacts"][0]
         assert "URI" in case_insensitive_recipe["Manifests"][0]["Artifacts"][0]
 
+    def test_read_json_with_syntax_err(self):
```
saranyailla, line 45: Nit: I know we've been naming the tests like this, but let's try making it more clear from next time onwards. 
For eg. We'd name this test to something like this: `Given_json_recipe_file_with_syntax_err_When_read_Then_throw_syntax_exception`

patrzhan: Fixed in latest commit

saranyailla (approved):

hancin (approved):

PR #249, opened 2023-09-27
chore: cleanup items for config update command and otf rename

patrzhan:
**Issue #, if available:**

**Description of changes:**
General cleanup of codebase prior to release. Includes:
- Internal renaming of OTF to GTF for variable names, etc. Some outward facing log outputs are updated too, mostly error logging.
- Moving the wizard folder under commands/config/update as all the classes are only relevant to that command. Rename all "wizard" references to "config" references.
- Tweak for the help message for test-e2e run command.
- Because codecov was reduced in an earlier PR to help move development forward, it is now raised back to its original (95%) level as necessary tests have been added.

Also, changes the config update command to no longer prompt for gdk version. The underlying code to support this is still included though, in case this becomes a relevant thing for future releases.

Keep the codebase in a state that reflects the direction the project wants to take regarding naming conventions.

Unit tests were updated and pass. Some manual testing done to ensure functionality is preserved.

**Checklist:**


saranyailla (approved):

tests/gdk/commands/test/config/test_InitConfiguration.py
```diff
@@ -152,7 +152,7 @@ def test_GIVEN_gdk_config_with_otf_version_WHEN_test_init_with_otf_version_arg_T
         )
         self.mocker.patch("gdk.common.configuration.get_configuration", return_value=config)
         init_config = InitConfiguration({"otf_version": "1.2.3"})
```
hancin, line 154: this file is still full of references to otf_version and otf in the test naming scheme, which I assume you are planning to fix?

uat/steps/test_e2e.py
```diff
@@ -10,14 +10,14 @@ def verify_test_files(context):
...
 
-@step("we verify that the OTF version used is {version}")
+@step("we verify that the GTF version used is {version}")
 def verify_test_framework_version(context, version):
     cwd = context.cwd if "cwd" in context else os.getcwd()
     build_system_file = Path(cwd).joinpath(GDK_TEST_DIR, "pom.xml")
 
     with open(build_system_file, "r") as f:
         content = f.read()
-        assert f"<otf.version>{version}-SNAPSHOT</otf.version>" in content, f"OTF version {version} is not used."
```
hancin, line 20: is it meant to still use the `otf.version` parameter?

patrzhan: Yup. That's actually still in the template, so it has to stay like that until the template gets updated.

tests/gdk/commands/test/config/test_InitConfiguration.py
patrzhan: The ones that were left were intentional, since they're referring to properties in the config file or arguments for the commands. We do still support "otf" values in these scenarios (to keep support for anyone still using the old values), and some of the tests actually directly test the scenario where both otf and gtf keys are provided and we verify that the gtf ones take precedence.

hancin (approved):

Issue #143, opened 2023-03-10
Zip file exclude file names that starts with test


cbenitez81:
Refer to https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/ba77abcd5c92dd222fa62ab957587d7edf5599d6/gdk/build_system/Zip.py#L84

```
if not options:
            ignore_list.extend([
                "test*",
                ".*",
                "node_modules",
            ])
        else:
            ignore_list.extend(options.get("excludes", []))
```

If a file starts with test, it gets excluded for the zipping. Many libraries have files that starts with that name, eg pandas and if we were to include the library as part of the component the program will fail. I think that this needs to be removed of clarified in the documentation

saranyailla:
Hi,
Thank you for reporting this issue. If you don't want gdk to exclude `test*` folders, please use `excludes` key under `options` entry of `build` in the `gdk-config.json` file to exclude the files you want to exclude from being zipped. 
If you don't specify this, gdk excludes `test*` by default.

This new feature is added from 1.2.0 onwards. More info in the [docs](https://docs.aws.amazon.com/greengrass/v2/developerguide/gdk-cli-configuration-file.html#gdk-config-format). 

```
      "build": {
        "build_system" : "zip",
        "options": {
           "excludes": [".*", "node_modules"]
        }
      }
```

cbenitez81:
I see the documentation but I don't see where is specified that `test*` is set by default, just in the code. Probably this needs to be more front and center or just leave these `[".*", "node_modules"]` by default.

cbenitez81:
I think that the proposed solutions does not work, I'm getting this error `Project configuration file 'gdk-config.json' is invalid. Please correct its format and try again. Error: Additional properties are not allowed ('excludes' was unexpected)` . My guess is that the schema in https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/5f320f60e3a9b7050d0b4ab472b62df815cad5f0/gdk/static/config_schema.json#L83 has typo and it should be "excludes" instead of "exclude"

saranyailla:
This is fixed in GDK v1.2.2.

cbenitez81:
I still think  that `test`  should not be included in the list by default. I have several projects that requires creating a virtual environment as part of the component that has pandas.testing, torch.testing, etc. The component breaks on arrival, it's a relatively obscure problem to figure it out. Any plans of fixing this issue?

[...]

PR #253, opened 2023-10-13
fix: only exclude files from root dir when building zip

patrzhan:
**Issue #, if available:**
https://github.com/aws-greengrass/aws-greengrass-gdk-cli/issues/143

**Description of changes:**
Only exclude excluded files from the zip build, if they are in the root directory of the build.

See issue for correspondence. Certain projects may need to include files in sub-directories that were incorrectly being excluded.

Manually tested using a sample project that had test* prefixed files in both root directory and sub-directories. Change correctly packages it such that the excluded files in root are removed but the ones not in root were kept.

Updated integration test for build command to verify that test* prefixed file in src subfolder is kept, but test* prefixed file in root folder is excluded during zip build.

**Checklist:**

### DIFF
--- a/gdk/build_system/Zip.py
+++ b/gdk/build_system/Zip.py
@@ -1,3 +1,4 @@
+import fnmatch
 import shutil
 import logging
 from pathlib import Path
@@ -38,10 +39,26 @@ class Zip(GDKBuildSystem):
             artifacts_zip_build = Path(zip_build).joinpath(utils.get_current_directory().name).resolve()
             utils.clean_dir(zip_build)
             logging.debug("Copying over component files to the '{}' folder.".format(artifacts_zip_build.name))
+
+            def ignore_patterns_in_root(rootdir, *patterns):
+                def _ignore_patterns_in_root(path, names):
+                    if str(path) == str(rootdir):
+                        ignored_names = []
+                        for pattern in patterns:
+                            ignored_names.extend(fnmatch.filter(names, pattern))
+                        return set(ignored_names)
+                    else:
+                        return set()
+                return _ignore_patterns_in_root
+
+            cwd = utils.get_current_directory()
             shutil.copytree(
-                utils.get_current_directory(),
+                cwd,
                 artifacts_zip_build,
-                ignore=shutil.ignore_patterns(*self.get_ignored_file_patterns(project_config)),
+                ignore=ignore_patterns_in_root(
+                                                cwd,
+                                                *self.get_ignored_file_patterns(project_config)
+                                              ),
             )
 
             # Get build file name without extension. This will be used as name of the archive.
--- a/gdk/build_system/Zip.py
+++ b/gdk/build_system/Zip.py
@@ -39,26 +39,21 @@ class Zip(GDKBuildSystem):
             artifacts_zip_build = Path(zip_build).joinpath(utils.get_current_directory().name).resolve()
             utils.clean_dir(zip_build)
             logging.debug("Copying over component files to the '{}' folder.".format(artifacts_zip_build.name))
+            root_directory_path = utils.get_current_directory()
 
-            def ignore_patterns_in_root(rootdir, *patterns):
-                def _ignore_patterns_in_root(path, names):
-                    if str(path) == str(rootdir):
-                        ignored_names = []
-                        for pattern in patterns:
-                            ignored_names.extend(fnmatch.filter(names, pattern))
-                        return set(ignored_names)
-                    else:
-                        return set()
-                return _ignore_patterns_in_root
+            def ignore_patterns_in_root(path, names):
+                if str(path) == str(root_directory_path):
+                    return {
+                        name
+                        for pattern in self.get_ignored_file_patterns(project_config)
+                        for name in fnmatch.filter(names, pattern)
+                    }
+                return set()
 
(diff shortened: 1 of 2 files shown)


integration_tests/gdk/components/test_integ_BuildCommand.py
```diff
@@ -26,8 +26,14 @@ def test_GIVEN_zip_build_system_WHEN_build_THEN_build_zip_artifacts(self):
         bc = BuildCommand({})
         bc.run()
         build_recipe_file = self.tmpdir.joinpath("greengrass-build/recipes/recipe.yaml").resolve()
+        included_file_path = "zip-build/" + self.tmpdir.name + "/src/test_do_want_this_file.txt"
+        excluded_file_path = "zip-build/" + self.tmpdir.name + "/test_dont_want_this_file.txt"
```
MikeDombo, line 30: use fstrings

patrzhan: done in new commit

MikeDombo (approved):

gdk/build_system/Zip.py
```diff
@@ -38,10 +39,26 @@ def build(self, **kwargs):
             artifacts_zip_build = Path(zip_build).joinpath(utils.get_current_directory().name).resolve()
             utils.clean_dir(zip_build)
             logging.debug("Copying over component files to the '{}' folder.".format(artifacts_zip_build.name))
+
+            def ignore_patterns_in_root(rootdir, *patterns):
```
parrottsquawk, line 43: What is the purpose of the nested method definition here? Focus on readability. It might work, but is there a simpler way that you can write it?

gdk/build_system/Zip.py
```diff
@@ -38,10 +39,26 @@ def build(self, **kwargs):
...
+                    if str(path) == str(rootdir):
+                        ignored_names = []
+                        for pattern in patterns:
+                            ignored_names.extend(fnmatch.filter(names, pattern))
+                        return set(ignored_names)
+                    else:
+                        return set()
+                return _ignore_patterns_in_root
+
+            cwd = utils.get_current_directory()
```
parrottsquawk, line 54: Recommend naming variables with a readable names rather than disemvoweling them or using abbreviations.

gdk/build_system/Zip.py
```diff
@@ -38,10 +39,26 @@ def build(self, **kwargs):
...
             shutil.copytree(
-                utils.get_current_directory(),
+                cwd,
                 artifacts_zip_build,
-                ignore=shutil.ignore_patterns(*self.get_ignored_file_patterns(project_config)),
+                ignore=ignore_patterns_in_root(
```
parrottsquawk, line 58: Can this be simplified instead of using * method?

gdk/build_system/Zip.py
```diff
@@ -38,10 +39,26 @@ def build(self, **kwargs):
...
             utils.clean_dir(zip_build)
             logging.debug("Copying over component files to the '{}' folder.".format(artifacts_zip_build.name))
+
+            def ignore_patterns_in_root(rootdir, *patterns):
+                def _ignore_patterns_in_root(path, names):
+                    if str(path) == str(rootdir):
+                        ignored_names = []
+                        for pattern in patterns:
+                            ignored_names.extend(fnmatch.filter(names, pattern))
+                        return set(ignored_names)
```
parrottsquawk, line 49: Here's a pythonic way of writing this same logic:
```
return {
    name
    for pattern in patterns
    for name in fnmatch.filter(names, pattern)
}
```

parrottsquawk (changes requested):

integration_tests/gdk/components/test_integ_BuildCommand.py
parrottsquawk: Why should we use fstrings instead of concatenation @patrzhan?

patrzhan: Both readability and so that we don't need to be worried about the type of the thing we are concatenating.

gdk/build_system/Zip.py
patrzhan: Changed in newest commit

gdk/build_system/Zip.py
patrzhan: Was to return the callable while keeping the same overall syntax/implementation as the shutils base implementation, but I've refactored it to be a lot more readable in the latest commit.

gdk/build_system/Zip.py
```diff
@@ -38,10 +39,21 @@ def build(self, **kwargs):
             artifacts_zip_build = Path(zip_build).joinpath(utils.get_current_directory().name).resolve()
             utils.clean_dir(zip_build)
             logging.debug("Copying over component files to the '{}' folder.".format(artifacts_zip_build.name))
+            root_directory_path = utils.get_current_directory()
```
MikeDombo, line 42: so this is going to do what I recommended to do (which is good of course :) ). However, it would be even better if we could still exclude things from subdirectories if we want to.

So I should be able to write a glob like `test/*` and also `*/test/*`

saranyailla (approved):

parrottsquawk (approved):

PR #254, opened 2023-10-19
feat: zip build exclusions support all globs

patrzhan:
**Issue #, if available:**
https://github.com/aws-greengrass/aws-greengrass-gdk-cli/issues/143
**Description of changes:**
Update zip build to accept full globs to match with when picking files/directories to exclude from the build. In GDK 1.4.0 and earlier, glob-like patterns were provided to match individual file names against, regardless of the subdirectory level. In PR: https://github.com/aws-greengrass/aws-greengrass-gdk-cli/pull/253, this was changed to only exclude items at the root level. This change supports full glob notation including directories, so that items that should only be excluded from root (i.e. `test*`) can do so, but items that should still be excluded from all subdirectories such as node_modules folders can also be excluded.

This also means the defaults have changed. The default globs for exclusion have been set to `test*`, `**/.*`, and `**/node_modules`, meaning that test* is matched only against the root directory, but hidden files and node_modules are excluded at any level by default. This also means that any custom patterns provided in the gdk-config.json file may need to be changed, as the behavior may have changed.

Bugfix for the linked issue while maintaining support for exclusion in any subdirectory level.

Manually tested changes with a sample project to confirm that expected directories and files were excluded based on various provided globs. Also updated unit/integration tests to test the functionality that is changed.

Since this feature impacts default values and may change behavior of existing projects that upgrade to this version, input is needed on how to mitigate impact if necessary.

**Checklist:**


gdk/common/utils.py
```diff
@@ -152,6 +154,15 @@ def is_recipe_size_valid(file_path):
     return file_size <= MAX_RECIPE_FILE_SIZE_BYTES, file_size
 
 
+def generate_ignore_list_from_globs(root_directory, globs):
```
saranyailla, line 157: Why is it not inside Zip? Who else is going to need this?

patrzhan: Moved it to Zip file

gdk/build_system/Zip.py
```diff
@@ -41,19 +43,11 @@ def build(self, **kwargs):
...
-                if str(path) == str(root_directory_path):
-                    return {
-                        name
-                        for pattern in self.get_ignored_file_patterns(project_config)
-                        for name in fnmatch.filter(names, pattern)
-                    }
-                return set()
-
-            shutil.copytree(
+            custom_copytree.copytree(
```
MikeDombo, line 46: why not keep using shutil, but with a custom filter that looks at the full path, rather than only the filename?

patrzhan: Yea at first I thought this wouldn't work but realized it would work after all. See latest commit.

MikeDombo (approved):

saranyailla (approved):

PR #256, opened 2023-10-24
feat: add smart warning if old excludes config is provided

patrzhan:
**Issue #, if available:**

**Description of changes:**
Adds a smart warning during component build that notifies user of the glob behavior change during zip exclusions if their provided excludes config uses exclusively the patterns that do not match to directories (old format). The warning also informs the user of how to achieve the old behavior by updating their config and specifies the exact change that needs to be made for this to happen.

The warning can be ignored by setting the GDK_EXCLUDES_WARN_IGNORE environment variable to true.

Example output (given ["node_modules"] as the excludes config):

`[2023-10-24 15:47:07] WARNING - Build option 'excludes' found in config and none of the provided patterns include directory matching. In GDK version 1.5.0, patterns for exclusions in zip builds were changed to use the glob format, so patterns with no specified directory pattern will only match at the root level of the project. If this is intentional, you can ignore this warning, but to achieve the old behavior excluding from each subdirectory, append '**/' to each pattern as such: ["**/node_modules"]. To ignore this warning in the future, set the GDK_EXCLUDES_WARN_IGNORE environment variable to true.`

Since there is a behavior change with how excludes patterns are handled, this change will notify users with a warning about the new behavior and provide steps to update their project if they desire the old behavior.

Manually tested. Also added integration tests to verify the warning exists if excludes is provided with old format patterns, but not when the env var is set.

**Checklist:**

### DIFF
--- a/gdk/build_system/Zip.py
+++ b/gdk/build_system/Zip.py
@@ -44,6 +44,7 @@ class Zip(GDKBuildSystem):
 
             unwanted_paths = self.generate_ignore_list_from_globs(root_directory_path,
                                                                   self.get_ignored_file_patterns(project_config))
+            self.smart_excludes_warning(project_config)
 
             def ignore_with_glob_support(dir, names):
                 ignore_set = set()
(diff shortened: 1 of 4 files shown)


gdk/build_system/Zip.py
```diff
@@ -126,3 +127,28 @@ def generate_ignore_list_from_globs(self, root_directory, globs):
...
+                          "were changed to use the glob format, so patterns with no specified directory pattern will " +
+                          "only match at the root level of the project. If this is intentional, you can ignore this " +
+                          "warning, but to achieve the old behavior excluding from each subdirectory, append \'**/\' " +
+                          "to each pattern as such: ")
+        GDK_EXCLUDES_ENV_KEY = "GDK_EXCLUDES_WARN_IGNORE"
+        build_options = project_config.build_options
+        excludes_list = build_options.get("excludes", [])
+        if not excludes_list or os.environ.get(GDK_EXCLUDES_ENV_KEY, "False").lower() == "true":
+            return
+        elif all(pattern.find("/") == -1 for pattern in excludes_list):
```
saranyailla, line 149: Are we assuming that if there's no`/` in the `excludes` list, then the customer is using `GDK<1.5.0`?

saranyailla: You can comment that here in the code for readability.

patrzhan: I've added this in the latest commit.

saranyailla (approved):

hancin (approved):

PR #259, opened 2023-11-10
feat: add recipe validation to build and publish commands

patrzhan:
**Issue #, if available:**

**Description of changes:**
Adds the Greengrass recipe schema to GDK. This is temporary until a more permanent solution to pull the schema down from a cloud source is added.

Add recipe schema validation to the component build and component publish commands. These will compare the project recipe file against the included Greengrass recipe schema and stop the command and produce a ValidationError if the recipe does not validate against the schema.

Validating component recipes against the schema is a planned feature and allows users of GDK to identify issues with their recipes earlier in the development process.

Manually tested. Added unit/integration tests for the additions.

Output of the build/publish commands may change depending on the cloud api implementation of schema validation. If changes need to be this output may be changed in a future PR.

**Checklist:**

### DIFF
--- a/gdk/commands/component/transformer/BuildRecipeTransformer.py
+++ b/gdk/commands/component/transformer/BuildRecipeTransformer.py
@@ -9,7 +9,7 @@ import gdk.common.consts as consts
 import gdk.common.utils as utils
 from gdk.commands.component.config.ComponentBuildConfiguration import ComponentBuildConfiguration
 from gdk.aws_clients.S3Client import S3Client
-from gdk.common.exceptions.error_messages import RECIPE_SIZE_INVALID, PROJECT_RECIPE_FILE_INVALID
+from gdk.common.exceptions.error_messages import RECIPE_SIZE_INVALID, PROJECT_RECIPE_FILE_INVALID, SCHEMA_FILE_INVALID
 
 
 class BuildRecipeTransformer:
@@ -42,6 +42,8 @@ class BuildRecipeTransformer:
             validator.validate_recipe(component_recipe.to_dict())
         except jsonschema.exceptions.ValidationError as err:
             raise Exception(PROJECT_RECIPE_FILE_INVALID.format(self.project_config.recipe_file, err.message))
+        except jsonschema.exceptions.SchemaError as err:
+            raise Exception(SCHEMA_FILE_INVALID.format(err.message))
 
         self.create_build_recipe_file(component_recipe)
 
--- a/gdk/commands/component/transformer/BuildRecipeTransformer.py
+++ b/gdk/commands/component/transformer/BuildRecipeTransformer.py
@@ -33,6 +33,7 @@ class BuildRecipeTransformer:
             raise Exception(RECIPE_SIZE_INVALID.format(self.project_config.recipe_file, input_recipe_file_size))
 
         component_recipe = CaseInsensitiveRecipeFile().read(self.project_config.recipe_file)
+        self.update_component_recipe_file(component_recipe, build_folders)
 
         logging.info("Validating the recipe against the Greengrass recipe schema.")
         try:
@@ -42,7 +43,6 @@ class BuildRecipeTransformer:
         except jsonschema.exceptions.ValidationError as err:
             raise Exception(PROJECT_RECIPE_FILE_INVALID.format(self.project_config.recipe_file, err.message))
 
-        self.update_component_recipe_file(component_recipe, build_folders)
         self.create_build_recipe_file(component_recipe)
 
     def update_component_recipe_file(self, parsed_component_recipe: CaseInsensitiveDict, build_folders):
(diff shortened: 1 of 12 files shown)


gdk/commands/component/transformer/BuildRecipeTransformer.py
```diff
@@ -31,6 +33,15 @@ def transform(self, build_folders):
             raise Exception(RECIPE_SIZE_INVALID.format(self.project_config.recipe_file, input_recipe_file_size))
 
         component_recipe = CaseInsensitiveRecipeFile().read(self.project_config.recipe_file)
+
+        logging.info("Validating the recipe against the Greengrass recipe schema.")
+        try:
+            recipe_schema_path = utils.get_static_file_path(consts.recipe_schema_file)
+            validator = RecipeValidator(recipe_schema_path)
+            validator.validate_recipe(component_recipe.to_dict())
+        except jsonschema.exceptions.ValidationError as err:
```
MikeDombo, line 42: bypass-able?

patrzhan: Not implemented to by bypassable, since the cloud behavior is going to be enforced and not opt-out

gdk/commands/component/transformer/BuildRecipeTransformer.py
```diff
@@ -31,6 +33,15 @@ def transform(self, build_folders):
             raise Exception(RECIPE_SIZE_INVALID.format(self.project_config.recipe_file, input_recipe_file_size))
 
         component_recipe = CaseInsensitiveRecipeFile().read(self.project_config.recipe_file)
+
+        logging.info("Validating the recipe against the Greengrass recipe schema.")
+        try:
+            recipe_schema_path = utils.get_static_file_path(consts.recipe_schema_file)
+            validator = RecipeValidator(recipe_schema_path)
+            validator.validate_recipe(component_recipe.to_dict())
```
MikeDombo, line 41: the recipe which is being validated, is it the recipe template written by the customer or is it the generated recipe after replacing the variables by GDK?

patrzhan: Currently the former, but the latter option seems to be the better option to test against (former behavior also causing the UATs to fail when the processes in the UAT should still result in a valid component build). Will update to validate the generated recipe as those are the values actually being used.

patrzhan: Updated in latest commit to validate the generated recipe after GDK replaces fields.

gdk/commands/component/transformer/BuildRecipeTransformer.py
```diff
@@ -32,6 +34,15 @@ def transform(self, build_folders):
 
         component_recipe = CaseInsensitiveRecipeFile().read(self.project_config.recipe_file)
         self.update_component_recipe_file(component_recipe, build_folders)
+
+        logging.info("Validating the recipe against the Greengrass recipe schema.")
+        try:
+            recipe_schema_path = utils.get_static_file_path(consts.recipe_schema_file)
+            validator = RecipeValidator(recipe_schema_path)
```
randolphtan121, line 41: This can also raise `jsonschema.exceptions.SchemaError` where the input schema does not abide by jsonschema standards. This should be accounted for in case the local recipe schema was touched accidentally. Will also be useful in the future to track if the schema we pull down had some unexpected issues.

patrzhan: Added handling for this and manually tested with an invalid schema

saranyailla: This more of a CLI/service error. It should be logged differently like this one here: https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/main/gdk/CLIParser.py#L198-L199. 

Also, add an integ test, just to validate the schema like we've for other schema here: https://github.com/aws-greengrass/aws-greengrass-gdk-cli/blob/main/integration_tests/gdk/common/test_integ_model_actions.py#L12

gdk/common/RecipeValidator.py
```diff
@@ -0,0 +1,27 @@
...
+import jsonschema
+
+
+class RecipeValidator:
+    def __init__(self, schema_file):
+        self._setup_schema(schema_file)
+
+    def validate_recipe(self, recipe):
+        processed_recipe = self._keys_to_lower(recipe)
+        jsonschema.validate(instance=processed_recipe, schema=self.schema, cls=jsonschema.validators.Draft7Validator)
```
saranyailla, line 11: Is it going to log lowercase or the original case of recipe keys?

randolphtan121 (approved):
Looks good to me, just make sure to follow up with Saranya's note on treating `SchemaError` as a CLI/service error

gdk/commands/component/transformer/BuildRecipeTransformer.py
patrzhan: Updated in latest commit. The error message now directs to the github issues page and added an additional integration test to validate the recipe schema.

saranyailla (approved):

PR #262, opened 2023-11-14
feat: add link to recipe reference in docs in error message if schema…

patrzhan:
… validation fails

**Issue #, if available:**

**Description of changes:**
Adds link to docs in output when schema validation fails.
Referring to the recipe docs may be helpful for customers facing recipes that are failing schema validation.
Manually tested that the error message looks good. Adjusted the tests that expect a different error message.

**Checklist:**

### DIFF
--- a/gdk/common/exceptions/error_messages.py
+++ b/gdk/common/exceptions/error_messages.py
@@ -7,12 +7,20 @@ PROJECT_RECIPE_FILE_NOT_FOUND = (
     "No valid component recipe is found. Please include a valid recipe file of the component to build with default."
 )
 PROJECT_CONFIG_FILE_INVALID = "Project configuration file '{}' is invalid. Please correct its format and try again. Error: {} "
-PROJECT_RECIPE_FILE_INVALID = "Project recipe file '{}' is invalid. Please correct its format and try again. Error: {} "
+PROJECT_RECIPE_FILE_INVALID = (
+    "Project recipe file '{}' is invalid. Please correct its format and try again.\nFor more information regarding " +
+    "component recipes refer to the docs here: https://docs.aws.amazon.com/greengrass/v2/developerguide/" +
+    "component-recipe-reference.html\nError: {} "
+)
 SCHEMA_FILE_INVALID = (
     "The Greengrass recipe schema file has an unexpected error.\nPlease report it at " +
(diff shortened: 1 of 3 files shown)


MikeDombo (approved):

gdk/common/exceptions/error_messages.py
```diff
@@ -7,12 +7,20 @@
...
 SCHEMA_FILE_INVALID = (
     "The Greengrass recipe schema file has an unexpected error.\nPlease report it at " +
     "https://github.com/aws-greengrass/aws-greengrass-gdk-cli/issues if the issue persists.\nError details: {} "
 )
-BUILD_RECIPE_FILE_INVALID = "Built recipe file '{}' is invalid. Please correct its format and try again. Error: {} "
+BUILD_RECIPE_FILE_INVALID = (
+    "Built recipe file '{}' is invalid. Please correct its format and try again.\nFor more information regarding " +
+    "component recipes refer to the docs here: https://docs.aws.amazon.com/greengrass/v2/developerguide/" +
+    "component-recipe-reference.html\nError: {} "
+)
```
saranyailla, line 23: Do we really need two different error messages? We can use the same message for both.

saranyailla (approved):

gdk/common/exceptions/error_messages.py
patrzhan: I just wanted to make it clear that it isn't the original recipe file in their project for the publish case.

MikeDombo: What's the difference between a "project" and "built" recipe file? How can one be valid and the other invalid?

patrzhan: The case I had in mind was if someone goes in the greengrass-build folder and edits the recipe to something that's invalid, then the publish command would be using an invalid recipe and so we go and validate it again. Shouldn't be a common case at all but if anything happens between the build and publish command the second validation during publish should still catch these errors.

MikeDombo: That seems unlikely. Providing the path to the file which was validated would be sufficient to clarify what's invalid.

patrzhan: Not a blocker, but a quick change so: https://github.com/aws-greengrass/aws-greengrass-gdk-cli/pull/263

PR #269, opened 2023-12-04
feat: get latest GTF release name from Github API

patrzhan:
**Issue #, if available:**

**Description of changes:**
Get the latest GTF release name from Github API and use it as the default value. If this API request fails (typically due to exceeding rate limits), fall bad on old behavior and use a hard-coded release value.

Getting the release dynamically should allow customers to benefit from GTF releases as soon as they are available, and lessen the need for us to release new GDK versions to followup GTF releases. The hard coded value will still need to be updated once in a while though.

Modified unit and integration tests that use the config value. Added new unit test for the new utils function. Manually tested as well.

**Checklist:**

### DIFF
--- a/gdk/common/GithubUtils.py
+++ b/gdk/common/GithubUtils.py
@@ -1,0 +1,10 @@
+import requests
+
+
+class GithubUtils:
+    def get_latest_release_name(self, owner, repository):
+        latest_release_api_url = f"https://api.github.com/repos/{owner}/{repository}/releases/latest"
+        response = requests.get(latest_release_api_url)
+        if response.status_code != 200:
+            response.raise_for_status()
+        return response.json()["name"]
--- a/gdk/common/GithubUtils.py
+++ b/gdk/common/GithubUtils.py
@@ -7,4 +7,4 @@ class GithubUtils:
         response = requests.get(latest_release_api_url)
         if response.status_code != 200:
             response.raise_for_status()
-        return response.json()["name"]
+        return response.json().get("name")  # We typically name our GTF releases by the version.
--- a/gdk/common/config/TestConfiguration.py
+++ b/gdk/common/config/TestConfiguration.py
@@ -23,8 +23,12 @@ class TestConfiguration:
         github_utils = GithubUtils()
         try:
             latest_gtf_version = github_utils.get_latest_release_name(GTF_REPO_OWNER, GTF_REPO_NAME)
-            self.gtf_version = latest_gtf_version
-            logging.info("Discovered %s as latest GTF release name.", self.gtf_version)
+            if latest_gtf_version is not None:
+                self.gtf_version = latest_gtf_version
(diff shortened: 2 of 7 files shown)


gdk/common/GithubUtils.py
```diff
@@ -0,0 +1,10 @@
+import requests
+
+
+class GithubUtils:
+    def get_latest_release_name(self, owner, repository):
+        latest_release_api_url = f"https://api.github.com/repos/{owner}/{repository}/releases/latest"
+        response = requests.get(latest_release_api_url)
+        if response.status_code != 200:
+            response.raise_for_status()
+        return response.json()["name"]
```
saranyailla, line 10: Don't retrieve name by the key without verifying or catching the KeyError.

gdk/common/GithubUtils.py
```diff
@@ -0,0 +1,10 @@
+import requests
+
+
+class GithubUtils:
+    def get_latest_release_name(self, owner, repository):
+        latest_release_api_url = f"https://api.github.com/repos/{owner}/{repository}/releases/latest"
```
saranyailla, line 6: This one should work even when the customer doesn't have git installed or signed into their account, correct?

gdk/common/GithubUtils.py
```diff
@@ -0,0 +1,10 @@
+import requests
+
+
+class GithubUtils:
+    def get_latest_release_name(self, owner, repository):
```
saranyailla, line 5: Nit: You can add a comment here that we typically name the testing framework release by its version.

gdk/common/config/TestConfiguration.py
```diff
@@ -14,6 +20,14 @@ def _set_build_config(self, test_build_config):
         self.test_build_system = test_build_config.get("build_system", self.test_build_system)
 
     def _set_gtf_config(self, test_config):
+        github_utils = GithubUtils()
+        try:
+            latest_gtf_version = github_utils.get_latest_release_name(GTF_REPO_OWNER, GTF_REPO_NAME)
+            self.gtf_version = latest_gtf_version
```
saranyailla, line 26: We should make this API call only if the customer hasn't set their GTF version. For eg, if customer sets the `gtf_version` to 1.3.0 in the config, we don't need to make this call. We could verify if that version exists and warn them if it doesn't.

gdk/common/GithubUtils.py
patrzhan: Yup, I'm not adding any authentication details to the request. It should work up until the rate limit, at which point the backup hardcoded value will have to be used.

gdk/common/config/TestConfiguration.py
patrzhan: In the future we will likely want to add an informational log line that notifies the user of GTF updates if they were to provide a version that isn't latest. To prepare for that addition, I'm running the API call every time.

saranyailla (approved):

hancin (approved):

PR #270, opened 2023-12-06
feat: add a suggestion to update GTF version if provided version is n…

patrzhan:
…ot latest

**Issue #, if available:**

**Description of changes:**
Adds a friendly suggestion to use the latest GTF version if a version provided in configuration is not the latest. Also provides a suggestion to use latest GTF if a version provided in command args is not the latest. However, the user will only see one of these suggestions and never see both.

Users may have old values in their configuration and not know when a new version is available. Similar if they are used to using a command with an older argument. This suggestion should help them use our latest versions.

Manually tested to see that the log appears correctly.

**Checklist:**

### DIFF
--- a/gdk/commands/test/config/InitConfiguration.py
+++ b/gdk/commands/test/config/InitConfiguration.py
@@ -18,15 +18,6 @@ class InitConfiguration(GDKProject):
             _version_arg = self._args.get("otf_version", None)
         if _version_arg:
             logging.info("Using the GTF version provided in the command %s", _version_arg)
-            if not self.test_config.upgrade_suggestion:
-                try:
-                    if Version(_version_arg) < Version(self.test_config.latest_gtf_version):
-                        logging.info(
-                            f"The current latest version of GTF is {self.test_config.latest_gtf_version}. Please consider "
-                            "using the latest version."
-                        )
-                except Exception as e:
-                    logging.debug("Not providing GTF update suggestion due to caught version error: %s", str(e))
             return self._validated_gtf_version(_version_arg)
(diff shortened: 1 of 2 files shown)


gdk/commands/test/config/InitConfiguration.py
```diff
@@ -17,6 +18,15 @@ def _get_gtf_version(self):
...
             logging.info("Using the GTF version provided in the command %s", _version_arg)
+            if not self.test_config.upgrade_suggestion:
+                try:
+                    if Version(_version_arg) < Version(self.test_config.latest_gtf_version):
+                        logging.info(
+                            f"The current latest version of GTF is {self.test_config.latest_gtf_version}. Please consider "
+                            "using the latest version."
+                        )
+                except Exception as e:
+                    logging.debug("Not providing GTF update suggestion due to caught version error: %s", str(e))
```
saranyailla, line 29: Nit: I think we can move this `try-except` under this `_validated_gtf_version()` method and remove the `upgrade_suggestion` boolean completely.

patrzhan: We will still need the boolean as the _validated_gtf_version function is called whether or not a command line arg was passed. However, I think it makes more logical sense to include the logic under that function and it is much more readable, and we can avoid the try-except logic as we'll already be validating the version beforehand. I'll move the logic over in the next commit and also renamed the boolean to be more clear what false/true means at a glance.

saranyailla (approved):

muboheng (approved):

